PR #14, opened 2016-11-20
some more examples for journals

Daniel-Mietchen:
added a few more examples from different fields, reordered alphabetically.

There is also a list of the journals most cited on the English Wikipedia: https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Academic_Journals/Journals_cited_by_Wikipedia/Popular1 .


fnielsen:
In production.

PR #19, opened 2016-12-09
slight rephrasing

Daniel-Mietchen:


fnielsen:
In production. Example https://tools.wmflabs.org/scholia/topic/Q52

PR #31, opened 2017-01-17
two more examples

Daniel-Mietchen:
also reordered by Q ID.


fnielsen:
Now in production https://tools.wmflabs.org/scholia/topic/

PR #92, opened 2017-03-30
changed default image

Daniel-Mietchen:
That 1D line image looks a bit odd, so I replaced it with a "no pic available" pic.

This could be further refined, e.g. by using https://commons.wikimedia.org/wiki/File:Replace_this_image_female_(red).svg or https://commons.wikimedia.org/wiki/File:Replace_this_image_male_(blue).svg if gender is female or male, or by using https://commons.wikimedia.org/wiki/File:No_portrait.svg and linking it to https://commons.wikimedia.org/wiki/Commons:Upload .


Daniel-Mietchen:
In principle, https://commons.wikimedia.org/wiki/File:No_image_available_600_x_200.svg would be better, but that long file name currently does not render properly in the image grid of the query results. Lots more placeholder images via https://commons.wikimedia.org/wiki/Category:Image_placeholders .

fnielsen:
The circle image seems a bit big to me. I wonder if we can find a smaller one.

fnielsen:
I have changed it to a white line which is "invisible".

fnielsen:
Another possibility could be a "person" icon with a 1x6 size.

PR #95, opened 2017-04-01
Restrict "publications" to 'scientific article's.

egonw:
By adding to the query:

      VALUES ?type { wd:Q13442814 }
      ?work wdt:P31 ?type .

It can be extended to other publications types (books), but would exclude others like data sets, conference posters, etc, which would be better in separate tables (see e.g. https://github.com/fnielsen/scholia/issues/94).


fnielsen:
I am somewhat reluctant. I think I would like to have books and these (for instance) there too, - untill I find a way to categorize the types of works? What do you think?

egonw:
The background is that I have quite some data sets, software releases, and other small things that will make the research papers unfindable... but I full see your perspective!

fnielsen:
A quickfix would be to display the "instance of" in a column.

egonw:
I like that idea... then people can sort.

fnielsen:
I am closing this for now. I have added a column with 44ca1bdcaaabb2e9286eb1cf124b32b64bf572a2 We might what to go back and see what is possible later.

PR #169, opened 2017-07-13
Added publications about the chemical

egonw:
Patch that adds two sections. The queries are slightly different from that of topic, as I specifically want to include works use as 'reference' of physchem properties.

Also fixed a duplicate < h2 > ID.

(I also note that that "data" and "field study" (e.g. for acetic acid) are not recognized as "works" and redirect to topic...)

### DIFF
--- a/scholia/app/templates/chemical.html
+++ b/scholia/app/templates/chemical.html
@@ -69,11 +69,43 @@ WHERE {
 } 
 ORDER BY ASC(?propEntityLabel)
 `
+  recentArticlesSparql = `
+SELECT ?date ?work ?workLabel ?type ?topics
+WITH {
+  SELECT DISTINCT ?work WHERE {
+    {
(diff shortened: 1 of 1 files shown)


egonw:
Oh, this is what it looks like for acetic acid:

image

fnielsen:
Why is/should is a paper with "DSSTox" listed in http://127.0.0.1:5000/chemical/Q159683 ? It comes from 

```
      wd:Q159683 ?propp ?statement .
      ?statement a wikibase:BestRank ;
                 prov:wasDerivedFrom/pr:P248 ?work .
```
And from the "DTXSID3020332" value on the https://www.wikidata.org/wiki/Q159683 page. Is that intended? 

I suppose it might be ok.

egonw:
It using the same approach as with listing the output of an author: we don't distinguish between types of output. The item on Figshare is the data set with InChIKey<>DSSTox ID mappings.

PR #174, opened 2017-07-16
Feature/pathway aspect

egonw:
cc @andrawaag

Patch introducing an aspect for biological pathways.

image


fnielsen:
Why would a biological pathway use P2860? And not P973?

fnielsen:
I have put up a question for Pathway bot https://www.wikidata.org/wiki/User_talk:Pathwaybot

PR #175, opened 2017-07-25
Added some Debian specific instructions

egonw:


fnielsen:
Why do you need "python3-flask". Does that  not come with the entry in `requirements.txt`?

egonw:
True. Debian packages of the modules come with their security support... that's all.

PR #178, opened 2017-08-04
Remove note from co-author graph

lucaswerkmeister:
The Wikidata Query Service UI, including the embed version, now retries queries with POST if the GET request fails because they’re too long (this was implemented in [Ic3ea53c50e][1]), so there is no hard limit on the number of authors. (The query sometimes fails for other reasons, but that can also happen with less than five authors.)

[1]: https://gerrit.wikimedia.org/r/#/c/355131/

### DIFF
--- a/scholia/app/templates/authors.html
+++ b/scholia/app/templates/authors.html
@@ -32,8 +32,6 @@
 
 <h2 id="Co-author graph">Co-author graph</h2>
 
-(Does not work with five or more authors)
-
 <div class="embed-responsive embed-responsive-4by3">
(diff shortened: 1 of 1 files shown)


fnielsen:
This is now running on Toolforge eg https://tools.wmflabs.org/scholia/authors/Q37876,Q173746

lucaswerkmeister:
Cool, thanks!

By the way, I just got an error even for those two authors – did you ever report that or shall I open a Phabricator ticket?

fnielsen:
Q37876,Q173746 does not make an error for me. What kind of error does it make?

lucaswerkmeister:
Just a generic “error” after a few seconds; in the network panel, I can see that the query service returns an exception – apparently there’s a `NullPointerException` somewhere in the GAS service. But it’s not reliably reproducible for some reason, I had to try a few combinations of items before I got it again.

fnielsen:
You better report that then, as I have never seen it.

lucaswerkmeister:
Done, https://phabricator.wikimedia.org/T172511.

PR #182, opened 2017-08-20
Added a section for articles citing a particular pathway (not too man…

egonw:
…y at this moment in Wikidata, but will grow; e.g. pathway Q30225468)

Current such citations: 

image

(Maybe wait until there is more such data?)

### DIFF
--- a/scholia/app/templates/pathway.html
+++ b/scholia/app/templates/pathway.html
@@ -53,9 +53,32 @@ ORDER BY DESC(?date)
 LIMIT 500
 `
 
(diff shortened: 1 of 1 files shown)


fnielsen:
I am still confused about the merging of reality with work.

egonw:
As in, does the entry represent the "concept" or a specific "work" or "the document describing the concept"? yeah, there are ontological wars about that... everyone understands the issues with any of the options, but there does not seem to be a correct answer...

PR #217, opened 2017-11-04
RSS Feature

egonw:
Adds RSS feeds for author, venue, and topic. This is a full draft and ready to be tested. If working, we can discuss adding RSS icons or links somewhere on the respective author, venue, and topic pages. Just add /latest/rss after /author/Qsomething...

For example:

image

### DIFF
--- a/scholia/rss.py
+++ b/scholia/rss.py
@@ -17,24 +17,65 @@ Examples:
 from __future__ import print_function
 
 import requests
+from dateutil.parser import parse as parse_datetime
+from email.utils import formatdate
+import time
+# from uuid import uuid3, NAMESPACE_URL
 
+SPARQL_QUERY = """
+PREFIX bd: <http://www.bigdata.com/rdf#>
+SELECT ?work ?workLabel ?date
+WITH {{
+  SELECT
+    (MIN(?dates) AS ?date) ?work
+  WHERE {{
+    ?work wdt:P50 wd:{q} .
+    ?work wdt:P577 ?datetimes .
+    BIND(xsd:date(?datetimes) AS ?dates)
+  }}
+  GROUP BY ?work ?workLabel
+  ORDER BY DESC(?date)
+  LIMIT 10
+}} AS %content {{
+  INCLUDE %content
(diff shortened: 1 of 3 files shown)


fnielsen:
I am wondering whether "latest" is sufficiently telling. Could one imaging a future latest-citations? What about latest-article/rss? Is that too long?

egonw:
It seems long... I have been thinking about this, and there is a reason why it has /latest/ ... but what other feeds to author, venue, and topic do we expect? (BTW, I think organization should have one too...)

fnielsen:
On Google Scholar I find that the citations to my articles are perhaps the most interesting. Would that be `latest-citations` in Scholia? 

For organizations: `latest-affiliated`?

egonw:
`latest-citations`++

But I am not sure what you mean with `latest-affiliated`?

scholia/rss.py
```diff
@@ -0,0 +1,212 @@
...
+  SERVICE wikibase:label {{ bd:serviceParam wikibase:language "[AUTO_LANGUAGE],en". }}
+}} ORDER BY DESC(?date)
+"""
+
+def _value(item, field):
+    return item[field]['value'] if field in item else ''
+
+def wb_get_author_latest_articles(q):
+    if not q:
+        return {}
```
fnielsen: Why is  a dictionary return here? Shouldn't this be an empty string?

fnielsen:
Sorry to bother you again. I changed my mind as wouldn't "work" be a better word than "articles". That will fit with the menu. 

I can change it.

fnielsen:
I am doing it now.

scholia/rss.py
egonw: Copy/paste issue... no particular reason.

egonw:
OK, I'll hold of hacking for now then... please ping me when done...

fnielsen:
Running now: https://tools.wmflabs.org/scholia/venue/Q6294930/latest-works/rss

scholia/rss.py
fnielsen: I now return an empty string instead in all 3 cases.

PR #231, opened 2017-12-24
Fixed the query to find literature about some chemical compound.

egonw:
It uses 'part of' but should have used 'has part', because the main subject should have a topic that has vanillin, and not be part of vanillin. The false positive was carbon, which is 'part of' vanillin.


egonw:
@fnielsen, OK, hold this off for now... there are more instances where this applies... I will add at least one more patch...

egonw:
@fnielsen, it seems the topic queries suffer from the same problem (which is possibly where I copied them from): uses P361... but maybe for topics it makes sense... sorry, my head is spinning now...

So, it was: articles about something that is "part of" the topic of interest... so, that goes wrongs for chemicals, where it lists tons of papers on carbon for any compound containing a carbon... but maybe that makes sense for topics? 

Say, any articles about me as "part of" Maastricht University should also be listed as article about Maastricht University? So, perhaps for the topic page it does make sense?

PR #235, opened 2017-12-29
Feature to view taxons

egonw:
image

### DIFF
--- a/scholia/app/templates/taxon.html
+++ b/scholia/app/templates/taxon.html
@@ -1,0 +1,63 @@
+{% extends "taxon_base.html" %}
+
+{% block scripts %}
+{{super()}}
+
+<script type="text/javascript">
+
+ variantsSparql = `
+SELECT ?variant ?variantLabel ?types
+WITH {
+  SELECT ?variant ?type
+  WHERE {
+    ?variant wdt:P3433 wd:{{ q }}
+  }
+} AS %variants
+WITH {
+  SELECT ?variant (GROUP_CONCAT(?type_label; separator=" // ") AS ?types) WHERE {
+    INCLUDE %variants
+    OPTIONAL {
+      ?variant wdt:P279 ?variantType . 
+      ?variantType rdfs:label ?type_label .  FILTER (lang(?type_label) = 'en')
+    }
+  }
+  GROUP BY ?variant
+} AS %result 
+WHERE {
+  INCLUDE %result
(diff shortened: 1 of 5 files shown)


fnielsen:
https://tools.wmflabs.org/scholia/taxon/Q15978631

fnielsen:
I am wondering whether we can use the space in the table columns to display something useful, e.g., an example publication with P921?

pigsonthewing:
Thank you. 

Unfortunately, this seems buggy. The parent taxa on https://tools.wmflabs.org/scholia/taxon/Q25485 do not match the one on https://www.wikidata.org/wiki/Q168473#P171

For example, Scholia has:

- Squamata | order
- Saurischia | order
- Passeriformes | order

while Wikidata only has the last of these.

fnielsen:
Well spotted! This is strange. I cannot see where the error is. Perhaps a Wikidata item further up in the hierarchy is linking it wrongly...

fnielsen:
http://tinyurl.com/y82crtk5  The bird class leads to a Coelurosauria suborder. That is wrong. It is through Paraves, Pennaraptora, Maniraptora, Maniraptoriformes. This will need to be fixed in Wikidata.

```
#defaultView:Graph
SELECT
  ?child ?childLabel
  ?parent ?parentLabel
WHERE {
  SELECT ?child ?childLabel ?parent ?parentLabel WHERE {
    service gas:service {
      gas:program gas:gasClass "com.bigdata.rdf.graph.analytics.BFS" ;
                  gas:in wd:Q168473 ; 
                  gas:traversalDirection "Forward" ;
                  gas:out ?child ;
                  gas:out1 ?depth ;
                  gas:out2 ?parent1 ;
                  gas:linkType wdt:P171 ;
                  gas:maxIterations 20;
                  gas:maxVisited 200 .
    }
    ?child wdt:P171 ?parent .
    ?child rdfs:label ?child_label . FILTER(LANG(?child_label) = 'en')
    ?parent rdfs:label ?parent_label . FILTER(LANG(?parent_label) = 'en')
    OPTIONAL {
      ?child wdt:P105 ?rank .
      ?rank rdfs:label ?rank_label_ . FILTER (LANG(?rank_label_) = 'en')
    }
    BIND(CONCAT(?child_label, " - ", COALESCE(?rank_label_, "???")) AS ?childLabel)
    BIND(CONCAT(?parent_label, " - ") AS ?parentLabel)
  }
}
```

fnielsen:
The above graph query is now part of the taxon aspect.

PR #236, opened 2017-12-29
Chemical classes are special and the regular chemical aspect does not…

egonw:
… work well. The new page looks like (except that "Related compound" in the screenshot is "Example compounds", capped at 500).

image

@fnielsen, please do take note of the change regarding figuring out what feature to show... for this patch I had to change the logic: if something cannot be types (P31), then it needs to determine the things the item is subclass of (more expensive) which is run only when no suitable aspect was found...

### DIFF
--- a/scholia/app/templates/chemclass.html
+++ b/scholia/app/templates/chemclass.html
@@ -80,6 +80,12 @@ LIMIT 500
 
 <div id="intro"></div>
 
+<h2 id="ClassHierarchy">Class Hierarchy</h2>
+
+<div class="embed-responsive embed-responsive-4by3">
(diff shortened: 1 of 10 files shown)


egonw:
Added a commit for a graphical visualization:

screenshot_20171231_130427

fnielsen:
Should we use a more fuller name than "chemclass"? For instance, "chemical-class". I have had a tendency to write the full version, e.g., "q-to-bibliography-templates". What do you think?

egonw:
Long URLs are hard to type, but since we have autocomplete and URL histories in browsers nowadays, I'm perfectly fine with this suggestion. I will make the updates asap!

egonw:
@fnielsen, I just pushed a commit to rename chemclass to chemical_class. Sorry for the delay.

egonw:
@fnielsen, I need your help here... when I use "chemical-class" is complains that show_chemical-class has a syntax error (which is why I went for all underscore earlier)... but when I keep using show_chemical_class but using chemical-class otherwise, I get an error that I want to use app.show_chemical-class even though that does not show up anywhere in the code, and seems dynamically generated by 'werkzeug'? (I did delete all .pyc files to make sure it was not some cache issue... or is it?)

fnielsen:
What I mean is that 'chemical-class' should be the variation seen by the user (in the URL or on the command-line) while chemical_class should be used in the code (because dash is not possible and that the underscore is standard Python style).

egonw:
OK, then I think we're talking about different things?

So,

* `scholia.query q-to-class $q` on the command line should answer with 'chemical-class';
* the URL should remain /chemical_class/; and 
* the code should remain using chemical_class?

fnielsen:
scholia.query q-to-class -> chemical-class
URL /chemical-class/
Code: chemical_class

fnielsen:
"scholia.query q-to-class -> chemical-class" does not work because it is directly used in the code. I have changed URL /chemical-class/ to eg "/chemical-class/Q41581", https://tools.wmflabs.org/scholia/chemical-class/Q41581

egonw:
No, classes don't have physchem props... maybe I was too focused on getting the page/URLs to work that I forgot to remove that??

egonw:
Sorry! Can you remove that part, or shall I make a PR?

egonw:
PR: https://github.com/fnielsen/scholia/pull/241

PR #256, opened 2018-02-25
URL pattern following current best practices

egonw:


fnielsen:
It is now running, e.g, https://tools.wmflabs.org/scholia/work/Q26899110. 

We probably need escapes for the DOI. I suppose there can be "<" in the DOI and many other bad things.

egonw:
True, about the "<"... my publist has such DOIs :)

PR #257, opened 2018-02-25
Show the ORCID for authors (if available)

egonw:
Which makes it look like:

image

@fnielsen, if you are OK, I can add an icon later...

### DIFF
--- a/scholia/app/templates/author.html
+++ b/scholia/app/templates/author.html
@@ -128,6 +128,7 @@ LIMIT 500
 
 <div id="intro"></div>
 
+<div id="details"></div>
 
 <h2>List of publications <a href="{{ url_for('app.show_author_empty') }}{{ q }}/latest-works/rss"><img height="24" width="48" src="{{ url_for('static', filename='images/rss-40674_320.png') }}" alt="RSS icon" /></a></h2>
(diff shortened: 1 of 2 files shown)


egonw:
Image patch pending....

egonw:
New layout:

image

fnielsen:
There is still an issue around the logo. I disabled it because of Toolforge rules/privacy.

egonw:
Oh, it's the icon being on a different server?

fnielsen:
Yes.

fnielsen:
Is there a copyright on the icon?

egonw:
Don't think so: https://orcid.org/trademark-and-id-display-guidelines

alainna:
CC-0 :) Let us know if you'd like it in another format!

egonw:
@alainna, I left a comment on Figshare... the icons there are square, but the Display Guidelines has round icons, which I actually like more... are those also available from Figshare?

alainna:
@egonw hum! The iD icons should download as a square, with the background displaying as transparent. Let me know if that's not the case and I'll a) provide direct links here, and b) update the Figshare listing 😄

PR #263, opened 2018-03-03
Chapters are also works

egonw:
Small patch, and maybe there are reasons why it is not like this? If so, please ignore.


fnielsen:
Now running, see, e.g., https://tools.wmflabs.org/scholia/author/Q20895241

PR #271, opened 2018-03-18
changing "persons" to "people" in events template

Daniel-Mietchen:


fnielsen:
There is something with this patch. I think you are reusing an old patch. Usually Egon makes new patches and names them according to what the patch does.

PR #274, opened 2018-03-23
Use the local ORCID icon

egonw:


scholia/app/templates/base.html
```diff
@@ -35,9 +35,9 @@
 
       try {
           var orcid = item.claims.P496[0].mainsnak.datavalue.value;
-          $( '#details' ).append( '<div><!-- a href="https://orcid.org/"><!-- img alt="' +
-	      'ORCID logo" src="https://orcid.org/sites/default/files/images/orcid_16x16.png"' +
-	      'width="16" height="16" hspace="4" /></a --> <a href="https://orcid.org/' +
+          $( '#details' ).append( '<div><a href="https://orcid.org/"><img alt="' +
+	      'ORCID logo" src="/static/images/orcid_16x16.gif"' +
```
fnielsen: I think there is a problem with "/..." due to tools.wmflabs.org having the /scholia prefix.

egonw: I copied that from another file, just to prevent issues like that :(

egonw: I'll push a fix...

egonw: testing the fix...

egonw:
@fnielsen, OK, better /static/ solution pushed... (as third commit to this PR)

PR #381, opened 2018-05-19
New layout for empty chemical / - class pages

egonw:
image


egonw:
And for classes no redirects (at this moment):

image

Issue #201, opened 2017-09-25
open disease, chemical, etc as "topic"


egonw:
Right now, some topics end up on a special page, like gene, disease, etc... but the "topic" page template often has additional, complementary information (queries). That made me think to have links/buttons just below the title with something like "open as topic"...

This "issue" is to discuss this idea...

egonw:
@fnielsen, would you be happy with just a link below the description (if available) saying "Open as topic"?

fnielsen:
Yes. There are a number of aspects that have this issue. For instance, some organizations can be viewed both as "organization" and "sponsor", - and I suppose also as topic.

egonw:
@fnielsen, you indicated this functionality was already implemented... so, can this issue be closed then?

fnielsen:
Yeah. The styling could be improved though and further catching

fnielsen:
Aspect-chooser: 4451de6725c9ff8295b0b4c0ae12d5a94fb02253

PR #383, opened 2018-05-19
Added 'Open as Topic' buttons (implements #201)

egonw:
Example:

image


fnielsen:
This is actually implemented

egonw:
Oh, I missed that... How?

fnielsen:
In JavaScript in the base.html
https://github.com/fnielsen/scholia/blob/2f6f7de7678694e62cd2394400b6abf042ec295d/scholia/app/templates/base.html#L107

I does not show the button if it is not a value used by P921.

The query could be extended and the UX could certainly also be improved.

PR #404, opened 2018-06-09
Also support P31 to select events of a series

nichtich:
At https://www.wikidata.org/wiki/Wikidata:WikiProject_Events the proposed property is P31 so this should be supported. I have not tried whether this may lead to timeouts.

### DIFF
--- a/scholia/app/templates/event_series.html
+++ b/scholia/app/templates/event_series.html
@@ -12,7 +12,10 @@ SELECT
   ?event ?eventLabel
   ?proceedings ?proceedingsLabel
 WHERE {
-  ?event wdt:P179 wd:{{ q }} .
+  {
+    { ?event wdt:P179 wd:{{ q }} } UNION
+    { ?event wdt:P31 wd:{{ q }} }
+  }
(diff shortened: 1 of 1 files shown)


fnielsen:
I wonder if you have some test cases?

fnielsen:
As far as I can see the timeline and the map has not been fixed.

nichtich:
The SPARQL query can also be simplified with `?event wdt:P179 | wdt:P31 wd:{{ q }}` instead of UNION.

fnielsen:
I have already changed from union to path queries: https://github.com/fnielsen/scholia/commit/2f6f7de7678694e62cd2394400b6abf042ec295d

However, there was one query were it did not work (line 63) which makes me wonder whether I have not understood path queries and unions enough.

nichtich:
My test case is https://tools.wmflabs.org/scholia/event-series/Q54856428 which has no proceedings so I did not test all queries. Putting parenthesis around all `(..|...)` may be more secure but I must admin it's only trial and error on my side.

PR #421, opened 2018-06-23
Use columns

egonw:
Small patch, but I like the new design.


fnielsen:
This is now running: https://tools.wmflabs.org/scholia/project/

PR #492, opened 2018-09-01
Added the RSS icon, redirecting to the RSS feed for the chemical as '…

### LINKED ISSUE #388: RSS button for /chemical/ Recent articles
Which is currently missing: 

image

egonw:
…topic' (closes #388)

@fnielsen, following your suggestion.


egonw:
Oh, one sec... I will do the same for /chemical_class/ ...

PR #562, opened 2018-12-02
Added the ACS

egonw:
Related to the petition to ask them to join the I4OC: https://www.change.org/p/asking-the-american-chemical-society-to-join-the-initiative-for-open-citations


fnielsen:
Now available in production: https://tools.wmflabs.org/scholia/publisher/

Issue #566, opened 2018-12-05
Replace links to new_resolve_authors with links to author-disambiguator


Daniel-Mietchen:
as per
https://github.com/fnielsen/scholia/search?q=new_resolve_authors&unscoped_q=sourcemd
and
https://tools.wmflabs.org/author-disambiguator/

i.e. instead of
https://tools.wmflabs.org/sourcemd/new_resolve_authors.php?doit=Look+for+author&name=Mary%20J%20Hamel
go for 
https://tools.wmflabs.org/author-disambiguator/?name=Mary+J+Hamel&doit=Look+for+author

PR #568, opened 2018-12-07
Link to Author Disambiguator instead of Author resolver

Daniel-Mietchen:
addressing #566.


Daniel-Mietchen:
I also removed mentions of @magnusmanske, since the tool is now developed collaboratively, and listing @arthurpsmith and other contributors properly would take more space than we have here.

Suggestions for better attribution are welcome.

Issue #547, opened 2018-11-08
more stats on /chemical/ front page


egonw:
like total number of chemicals, the number of chemicals with classification.

More ideas?

egonw:
- [ ] number of chemicals with more than 5 articles
- [ ] number of chemicals with more 1 or more articles
- [ ] number of chemicals with a boiling point
- [ ] number of chemicals with a melting point
- [ ] number of chemicals with a pKa

PR #569, opened 2018-12-08
More statistics about chemicals (implements #547)

egonw:
Perhaps too slow for your taste? The two queries with links to literature take 1.5~ish secs each... but it cached (for a short time).

### DIFF
--- a/scholia/app/templates/chemical_empty.html
+++ b/scholia/app/templates/chemical_empty.html
@@ -71,14 +71,45 @@ Chemical substances.
 
  statisticsSparql = `
 
+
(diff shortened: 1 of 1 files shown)


Daniel-Mietchen:
I just ran the query a few times and find the speed acceptable.

fnielsen:
I suppose the question of whether it is acceptable for use is less relevant than whether it is acceptable to WDQS. I am worried to make any SPARQL on the main page too complex. I suppose that not too many go to the aspect index (empty) pages, so it is less of a problem. After all I had no quarrels putting in a complex query on the project index page.

PR #592, opened 2018-12-23
Use AUTO_LANGUAGE everywhere

egonw:


egonw:
I was mostly concerned with the redundancy... can we make this a global value and then insert it into all the SPARQL queries, so that we have one place to add/update it? So the SPARQL query would have something like `{{ q }}` in the Scholia source pages, e.g.:

```SPARQL
 topicsSparql = `
#defaultView:Table
SELECT ?count ?theme ?themeLabel ?example_work ?example_workLabel
WITH {
  SELECT (COUNT(?work) AS ?count) ?theme (SAMPLE(?work) AS ?example_work)
  WHERE {
    ?work wdt:P50 wd:{{ q }} .
    ?work wdt:P921 ?theme .
  }
  GROUP BY ?theme
} AS %result
WHERE {
  INCLUDE %result
  {{ serviceLabel }}
}
ORDER BY DESC(?count) 
```

And that new variable is globally defined somewhere, either `SERVICE wikibase:label { bd:serviceParam wikibase:language "en,da,de,es,fr,jp,no,ru,sv,zh". }` or `SERVICE wikibase:label { bd:serviceParam wikibase:language "[AUTO_LANGUAGE],en". }`.

What do you think?

Daniel-Mietchen:
Instead of constructs like `{{ q }}` in the Scholia source pages, I would prefer us to 
- have the SPARQL query stored in plain, perhaps with a `?default` variable. if we go that way, we could even think about storing the queries using a specific property like Wikidata SPARQL query equivalent ([P3921](https://www.wikidata.org/wiki/Property:P3921)), which would open up new ways to configure Scholia page content (e.g. in connection with ASK queries, as per #617 )
- specify elements of that default SPARQL query that Scholia then works with to display content for the target QID(s) (e.g. replacing the `?default` variable with the target QID)

PR #594, opened 2018-12-23
Added a redirect based on a PubMed identifier

egonw:
Looks like (also as CLI):

image


Daniel-Mietchen:
Looks good to me except that I'd use another example &mdash; the profile at 
https://tools.wmflabs.org/scholia/work/Q42371516
is not very rich, so I suggest using one of those listed at https://www.wikidata.org/wiki/User:Daniel_Mietchen/Scholia/Examples/works .

PR #595, opened 2018-12-23
Fixed linking out to JSmol for some InChIs (fixes #391)

### LINKED ISSUE #391: ID's should be escaped before turned to URIs
For, e.g., https://tools.wmflabs.org/scholia/chemical/Q27124756 the InChl external link does not resolve because the ID is not escaped. This is also a security issue.

I cannot find an escaping function in Blazegraph's SPARQL.

egonw:
I think further replacement may me needed, and this approach can also be used for other properties (think SMILES) later.

### DIFF
--- a/scholia/app/templates/chemical.html
+++ b/scholia/app/templates/chemical.html
@@ -19,13 +19,13 @@ SELECT ?IDpred ?IDpredLabel ?id (SAMPLE(?idUrls) as ?idUrl) WITH {
   { SELECT * WHERE {
       INCLUDE %RESULTS
       FILTER (?IDpred = wd:P234)
-      BIND(IRI(REPLACE(?formatterurl, '\\$1', REPLACE(REPLACE(?id, "\\+", '%2B'), '/', '%2F'))) AS ?idUrls).
+      BIND(IRI(REPLACE(?formatterurl, '\\\\$1', ENCODE_FOR_URI(str(?id)))) AS ?idUrls).
     }
   } UNION {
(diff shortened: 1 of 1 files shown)


scholia/app/templates/chemical.html
```diff
@@ -16,8 +16,18 @@
     }
   } LIMIT 500
 } AS %RESULTS {
-  INCLUDE %RESULTS
-  BIND(IRI(REPLACE(?formatterurl, '\\\\$1', str(?id))) AS ?idUrls).
+  { SELECT * WHERE {
+      INCLUDE %RESULTS
+      FILTER (?IDpred = wd:P234)
+      BIND(IRI(REPLACE(?formatterurl, '\\$1', REPLACE(REPLACE(?id, "\\+", '%2B'), '/', '%2F'))) AS ?idUrls).
```
fnielsen: Would `ENCODE_FOR_URI` be better here? 

```SPARQL
SELECT * WHERE {
  BIND(ENCODE_FOR_URI("1S/C6H8O6/c7-1-2(8)5-3(9)4(10)6(11)12-5/h2,5,7-8,10-11H,1H2/t2-,5+/m0/s1") AS ?link)
}
```

egonw: I'll test that today! Definitely better code :)

egonw: Plz try again.

fnielsen:
Example: http://127.0.0.1:8100/chemical/Q27124756

PR #654, opened 2019-03-03
More chemical classes nowadays

egonw:


fnielsen:
Example: https://tools.wmflabs.org/scholia/Q161647

PR #699, opened 2019-05-10
Recognize academic chapters as works

egonw:


fnielsen:
An example would be https://tools.wmflabs.org/scholia/Q43149096

fnielsen:
Now running

PR #726, opened 2019-05-26
removing the OPTIONAL for dates in query for recent publications

Daniel-Mietchen:


fnielsen:
What was the rational for this change?

Daniel-Mietchen:
"Recent" implies some sort of time, but that does not make much sense if P577 is not set. Plus, for people with more than 10 publications, those entries without P577 would not show up in the default view anyway. 

The trigger for me to look into this was #727 (i.e. oldest publications).

PR #735, opened 2019-05-30
Added a section with highly cited articles with no (0) or little (1) …

egonw:
…main subject annotations

I also added only 1 topic because other sections need two topics, e.g. the topics graph.


Daniel-Mietchen:
Looks good to me but could be improved by actually listing the existing topic tag (and its label) for the 1-topic case.

egonw:
Yes, that should be doable.

Issue #738, opened 2019-06-05
Duplicate info in Scholia work


GerardMeijssen:
in https://tools.wmflabs.org/scholia/work/Q30512726 in het kopje "Supports the following statement(s)".. duplicatie van data; een keer 0000-0002-3200-1601 en een keer https://orcid.org/0000-0002-3200-1601

egonw:
Thanks for reporting, and apologies I did not have time to look at it yet.

egonw:
This is what it looks like:

![image](https://user-images.githubusercontent.com/26721/59346212-14d5e280-8d12-11e9-9a24-e9590c99e593.png)

Daniel-Mietchen:
Looks like we have to use ``wikibase:statementValueNormalized`` to get at the normalized values, as per https://www.mediawiki.org/wiki/Wikibase/Indexing/RDF_Dump_Format .

egonw:
@Daniel-Mietchen, has that patch been applied yet?

Daniel-Mietchen:
No, that patch has not been applied yet, and I think there was an issue with it but don't remember, so just bumping this up for now.

egonw:
> and I think there was an issue with it but don't remember

The problem is that it's tricky. You cannot ask for just the normalized values, because that triple is missing when there is nothing to normalize.

I think this SPARQL fragment should do it:

```sparql
    INCLUDE %statements
    ?item ?p ?statement .
    ?property wikibase:claim ?p . OPTIONAL { ?property wikibase:statementValueNormalized ?a1 }
    ?statement ?a1 ?value1 ; ?a2 ?value2 . BIND (COALESCE(?value1, ?value2) AS ?value) BIND (COALESCE(?a1, ?a2) AS ?a)
    FILTER (!STRSTARTS(LCASE(STR(?value)), "http://wikiba.se/ontology#"))
    FILTER (!STRSTARTS(LCASE(STR(?value)), "http://www.wikidata.org/value/"))
    FILTER (!STRSTARTS(LCASE(STR(?value)), "http://www.wikidata.org/reference/"))
    OPTIONAL {?statement ?psv_statement_predicate ?psv_statement .
    ?statement_predicate_property wikibase:statementValue ?psv_statement_predicate.
    ?psv_statement wikibase:quantityUnit ?unit}
```

PR #740, opened 2019-06-13
fix for #738

Daniel-Mietchen:


Daniel-Mietchen:
Oops - that was a bit too quick.

PR #823, opened 2019-09-13
Additional properties and interactive views for pathways

AlexanderPico:
- Introduces two new sections: An interactive pathway viewer iframe and a link to organism
- Utilizes a new generic sparql query that is extensible for future uses
- Fails gracefully by hiding sections if data is not available
- Adds `const` per convention

Please let us know if we can improve upon this PR in any way. We are eager to contribute to scholia for pathway views.

### DIFF
--- a/scholia/app/templates/pathway.html
+++ b/scholia/app/templates/pathway.html
@@ -5,14 +5,6 @@
 
 <script type="text/javascript">
     
-    speciesSparql = `
-  SELECT $organism ?organismLabel 
-    WHERE {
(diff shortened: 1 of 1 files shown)


AlexanderPico:
Here are example pages that will be improved by this PR (see screenshot):

- https://tools.wmflabs.org/scholia/pathway/Q30151254
- https://tools.wmflabs.org/scholia/pathway/Q29892242

Screen Shot 2019-09-15 at 11 38 04 AM

Issue #61, opened 2017-03-14
UTF-8 problem with bibtex


fnielsen:
Scholia generates bibtex files with UTF-8 characters that bibtex cannot understand. Bibtex's sorting and author name extraction fails. Either users are forced over to biber or something must be done, e.g., a conversion table and accent conversion.

egonw:
Suggestion: export [BiBLaTeX](https://tex.stackexchange.com/questions/13509/biblatex-in-a-nutshell-for-beginners) which does export Unicode.

fnielsen:
Isn't there a problem with the backend not having good support of styles?

egonw:
as in CSL?

larsgw:
I can work on the accents, I just did some work on that for my new BibTeX parser.

PR #841, opened 2019-10-06
Support UTF8 in TeX output

larsgw:
- [x] accents
  - [ ] other symbols
  - [ ] tests

  1. One thing I found a bit confusing is that even with the accent conversion working (i.e., the BibTeX output contains just `{\. s}`, pdfLaTeX on Overleaf still complains about an unknown Unicode character, *with* `\usepackage[utf8]{inputenc}`.
  2. I'm not really familiar with Unicode strings in Python 2.7 yet, so there may be some problems there.
  3. For other symbols, there's [a great, big list](https://www.rpi.edu/dept/arc/training/latex/LaTeX_symbols.pdf) and I don't really know if we want to include all of that.

See #61

### DIFF
--- a/scholia/tex.py
+++ b/scholia/tex.py
@@ -32,6 +32,7 @@ from __future__ import print_function
 from os.path import splitext
 
 import re
+import unicodedata
 
 from six import u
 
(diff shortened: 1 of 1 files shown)


egonw:
@larsgw, I have been using Python3. I think Python2 was never really good at Unicode.

fnielsen:
There was a Python2 Unicode doctest issue that I have fixed. I wonder if the error message showed up in your `tox`?

larsgw:
I think so, yes.

PR #848, opened 2019-10-08
added author-award table for venues (closes #713)

### LINKED ISSUE #713: In journal aspect, consider adding a panel for awards received by people who published in there
[Prototype](https://query.wikidata.org/#SELECT%20DISTINCT%20%3Fauthor%20%3FauthorLabel%20%3Faward%20%3FawardLabel%20WHERE%20%7B%0A%20%20%3Fitem%20wdt%3AP1433%20wd%3AQ15763595%20%3B%0A%20%20%20%20%20%20%20%20wdt%3AP50%20%3Fauthor%20.%0A%20%20%3Fauthor%20wdt%3AP166%20%3Faward%20.%0ASERVICE%20wikibase%3Alabel%20%7B%20bd%3AserviceParam%20wikibase%3Alanguage%20%22%5BAUTO_LANGUAGE%5D%2Cen%22.%20%7D%20%20%0A%7D): 

```SPARQL
SELECT DISTINCT ?author ?authorLabel ?award ?awardLabel WHERE {
  ?item wdt:P1433 wd:Q15763595 ;
        wdt:P50 ?author .
  ?author wdt:P166 ?award .
SERVICE wikibase:label { bd:serviceParam wikibase:language "[AUTO_LANGUAGE],en". }  
}
```

mkutmon:


fnielsen:
I am wondering if we could make it a bit more informative by count, date or description?

fnielsen:
I think we also need a LIMIT since, e.g., Q180445 (Nature) times out.

fnielsen:
Now running, e.g., https://tools.wmflabs.org/scholia/venue/Q6294930

scholia/app/templates/venue.html
```diff
@@ -352,5 +361,9 @@ <h3>Authorships</h3>
 <table class="table table-hover" id="authorships-gender-distribution"></table>
 
 
+<h2 id="AuthorAwards">Author Awards</h2>
```
fnielsen: I am wondering whether there should be a space in "AuthorAwards"?

scholia/app/templates/venue.html
```diff
@@ -240,6 +240,14 @@
  ORDER BY DESC(?count)
  `
 
+ authorAwardsSparql = `
+SELECT DISTINCT ?author ?authorLabel ?award ?awardLabel WHERE {
+  ?item wdt:P1433 wd:{{ q }} ;
+        wdt:P50 ?author .
+  ?author wdt:P166 ?award .
+SERVICE wikibase:label { bd:serviceParam wikibase:language "[AUTO_LANGUAGE],en". }
```
fnielsen: I extended this with some more languages. Maybe the service labeling should be parametrized.

scholia/app/templates/venue.html
mkutmon: That's just the id of the header, right? The text itself has a space.

mkutmon:
> I am wondering if we could make it a bit more informative by count, date or description?

I'll check when I have time. We should be able to group the information nicely.

fnielsen:
I have already added the information.

scholia/app/templates/venue.html
fnielsen: Yes. On the other hand, I do not know if it is a good idea to have spaces in the ids (as I have been doing in the past). Do you know if one can run into problems with that? Maybe it is better with a dash?

mkutmon: Ah okay. I never used spaces in ids. 
Just found this for example: "HTML 5 is even more permissive, saying only that an id must contain at least one character and may not contain any space characters."

So I guess it's better to remove spaces. It doesn't really matter if it's author-awards or AuthorAwards or authorAwards but I would try to stick to one naming convention.

fnielsen: Thanks for the lookup. Maybe we should convert the spaces to a dash and lowercase when the word is not the first, e.g., "Author-awards". I hope that does not collide with the table-id.

Issue #721, opened 2019-05-21
Matrix visualization


fnielsen:
A matrix visualization has been setup now, see particularly now: https://github.com/fnielsen/scholia/commit/f0064a00602df9e6690addccd9a0f2bac74747a1

Example is running on author pages as "topics-works matrix", example: https://tools.wmflabs.org/scholia/author/Q41533116

There are some issue related to this:
- Clean up code, e.g., move it to a general sparqlToMatrixVisualization function in scholia.js.
- Right-most field might not give a tooltip.
- Resizing window does not seem to work.
- Problem with few (non or one) works/topics that does not render that well.
- Mobile phone issue. The visualization is not that suitable for a phone

fnielsen:
Phabricator: https://phabricator.wikimedia.org/T223809

fnielsen:
* If the visualization area is too small then it seems that the tooltip does not show.

Issue #69, opened 2017-03-16
Implement Javascript more


fnielsen:
Implement Javascript more:

- Display property values from the item itself.
- Make better more interactive results formats, e.g., with links from labels to other scholia pages

dartar:
Thanks for the pointer, the title is a misnomer indeed :) I'd focus more on describing user stories and user value and less on implementation to give people a chance to discuss design/feature requirements.

PR #852, opened 2019-10-09
Add conference poster (wd:Q54670950) to co-author graph query in Organization aspect

cthoyt:
It's pretty hard to see the diff, but I just added "%20wd%Q54670950" after the previous item "%20wd%3AQ947859"


fnielsen:
Do we have a test example organization? I am wondering whether https://tools.wmflabs.org/scholia/organization/Q1451981 would do?

cthoyt:
Yes - if the query is working correctly (I actually tested this query on Fraunhofer, since it is my workplace) we should see an edge appear between my name and Marc Zimmerman (I did indeed see this).

I could try writing a unit test for this as well. Was there a reason why some queries were written as javascript and some were made with iframes? I would prefer the javascript because they could also be handled as templates

fnielsen:
> Was there a reason why some queries were written as javascript and some were made with iframes? I would prefer the javascript because they could also be handled as templates

Yes. We have not yet programmed the plots with Javascript (except for the work-topic matrix), so we rely entirely on WDQS' plots, - which is suboptimal. This has been up for some time, see #69. The matrix visualization from May 2019 is the only one non-WDQS visualization we implemented so far, - and event that could be improved #721.

fnielsen:
I get "jQuery.Deferred exception: malformed URI sequence"

fnielsen:
it think "%Q54670950" is the problem. You need an escaped colon.

fnielsen:
The patch name could be more telling.

cthoyt:
Sorry about the patch name - I made it from the GitHub web interface and it's not so obvious how to change. Thanks for the fix!

fnielsen:
This should be working now, e.g., https://tools.wmflabs.org/scholia/organization/Q1451981

PR #875, opened 2019-10-22
Adds description, authors and ontologies where available

AlexanderPico:
Nice display of newly added information to hundreds of pathway items (see screenshot). New divs are hidden if no information is available.

Here is an example item with relevant content: https://tools.wmflabs.org/scholia/pathway/Q66104607

Screen Shot 2019-10-22 at 12 06 58 PM


egonw:
One comment for next time: best to work in branches and not in `master`.

scholia/app/templates/pathway.html
```diff
@@ -77,11 +77,13 @@
 
 //Extensible query for muliple, optional values
 const optionalDataValuesSparql =`
-SELECT ?wpid ?organism ?organismLabel
+SELECT ?wpid ?organism ?organismLabel ?description
```
fnielsen: Why not ?organismDescription  with automatic description?

scholia/app/templates/pathway.html
```diff
@@ -86,8 +86,33 @@
   }
 }
 `
+const  listOfAuthorsSparql = `
```
fnielsen: Should this maybe just be called authorsSparql so there is a one-to-one correspondence with the Javascript name in the HTML ID?

AlexanderPico:
Thanks for the great feedback! I will make another pass on the code and submit a new PR as a branch with the suggested changes. Feel free to close this one.

scholia/app/templates/pathway.html
AlexanderPico: Ah, we are not getting the organismDescription here. We are getting the description of the item `wd:{{q}} schema:description ?description`.  Is there a better way to retrieve that?

scholia/app/templates/pathway.html
AlexanderPico: Fixed in next PR

AlexanderPico:
See updated patch here: https://github.com/fnielsen/scholia/pull/877

Issue #850, opened 2019-10-08
Add retraction query to journal or author pages


fehrhart:
A query that can retrieve retracted articles e.g.
- list the retracted articles of a specific journal (maybe add to journal pages)
- list the retracted articles of a specific author (maybe add to author pages - discuss first!)
- list the retracted articles that are cited by other articles in that journal
- list the retracted articles that are cited by other articles

egonw:
Here's the query that lists articles for a journal (Nature as example) that cite retracted articles:

```sparql

SELECT DISTINCT
  ?retracted_work ?retracted_workLabel
  ?date
  ?citing_work ?citing_workLabel
WITH {
  # Find retracted papers indicated by instance of retracted paper, 
  # by retraction notice property or by significant event
  SELECT DISTINCT ?retracted_work WHERE {
    { ?retracted_work wdt:P31 wd:Q45182324 }
    UNION
    { ?retracted_work wdt:P5824 [] }
    UNION
    { ?retracted_work wdt:P793 wd:Q45203135 }
  }
} AS %retracted_works
WHERE {
  INCLUDE %retracted_works
  ?citing_work wdt:P2860 ?retracted_work ; wdt:P1433 wd:Q180445 .
  OPTIONAL {
    ?retracted_work wdt:P5824 ?retraction .
    ?retraction wdt:P577 ?retraction_datetime
    FILTER ( ?citing_work != ?retraction )
  }
  MINUS { ?citing_work wdt:P31 wd:Q1348305 }
  OPTIONAL {
    ?retracted_work p:P793 [ ps:P793 wd:Q45203135 ; pq:P585 ?keyevent_datetime ]
  }
  BIND(COALESCE(xsd:date(COALESCE(?retraction_datetime, ?keyevent_datetime)), "Unknown date") AS ?date)
  SERVICE wikibase:label { bd:serviceParam wikibase:language "[AUTO_LANGUAGE],en". }
}
ORDER BY DESC(?date)
```

fehrhart:
Done by #914

fnielsen:
Why is this marked done? The pull requests that I examine stille has issues?

fnielsen:
I am looking at 185347604a316d72cc8f64fa626d6aed42ed52b2

carlinmack:
Closing as #914, #1453 and #1466 are now merged

PR #876, opened 2019-10-23
Fehrhart patch retractions

fehrhart:
Please check, had some troubles with testing!
#850


fnielsen:
# It is a bit easier to overlook the changes in GitHub if there is a single commit rather than the multiple commits (though `git diff master` seems to display all changes fine.
# The headline "Retracted papers in this journal" seems to be wrong because it is for a publisher, - not a journal
# The Javascript name and HTML IDs should correspond (Javascript name should be lower case too)

scholia/app/templates/publisher.html
```diff
@@ -142,6 +174,16 @@ <h3 id="As function of number of published works">As function of number of publi
 </div>
 
 
+<h2>Retractions</h2>
+
+<id="Retractions">Retracted papers in this journal</h2>
```
egonw: Change to "Retracted papers in journals from this publisher"?

scholia/app/templates/publisher.html
```diff
@@ -142,6 +174,16 @@ <h3 id="As function of number of published works">As function of number of publi
...
+
+<table class="table table-hover" id="retracted"></table>
+
+<div class="embed-responsive embed-responsive-4by3">
```
egonw: @fehrhart, why are lines 183-185 with the `iframe` here? If I follow the URL, I just get the WDQS. What are you trying to do here?

fehrhart:
Testing now works for me.

One problem, the query works only on journals but not on publishers. Do we have a template for (scientific) journals? I could not find something suitable in the template folder.

fehrhart:
Working on a different patch in the correct template.

PR #877, opened 2019-10-23
Wikipathways patch1: adds description, authors and ontologies to pathway template

AlexanderPico:
Redo of https://github.com/fnielsen/scholia/pull/875, taking advice for branching and sqarql query constant name.


AlexanderPico:
Ah, and here is an example pathway item that will include authors, description and ontologies with this patch: https://tools.wmflabs.org/scholia/pathway/Q30225384

PR #878, opened 2019-10-24
changing the text for the leftmost menu button from "Scholia" to "Home"

Daniel-Mietchen:
to improve user navigation, as per https://en.wikipedia.org/w/index.php?title=Template_talk:Scholia&oldid=922620697 .

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -206,7 +206,7 @@
 {% block navbar %}
 <div class="navbar navbar-static-top">
     <ul class="nav nav-pills">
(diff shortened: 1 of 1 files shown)


fnielsen:
One is that with this change we do not have any indication of that this is Scholia.

egonw:
We can also replace the text with the logo?

Daniel-Mietchen:
Yes, replacing this with the logo would make sense.

Daniel-Mietchen:
This is now superseded by #956 .

Issue #302, opened 2018-04-01
Try to catch the "Rate limit exceeded" error and display something more useful


Daniel-Mietchen:
Sometimes, loading a Scholia aspect yields a 
```
Rate limit exceeded
```
message for one or more of the panels.

The problem is typically resolved by just reloading the page a few moments later but there may be something we can do to reduce the likelihood of that to happen, e.g. by adding a small random delay between the SPARQL calls for a Scholia page.

fnielsen:
Another way would be to delay the embedding, only display the component when the user clicks

Daniel-Mietchen:
The reason for this is that too many requests are being made within a certain time frame.

Several options to reproduce &mdash; key thing is to run multiple different queries at overlapping times.

For instance, you could

1. choose a panel with lots of links to richly annotated topics, e.g. https://scholia.toolforge.org/venue/Q2000008#topics-header
1. quickly right-click on multiple Scholia links (e.g. the entries in the Topics column in the example table) to open new tabs with Scholia profiles for those topics.
1. check these topic profile pages for Server errors, some or all of which may come with a "Rate limit exceeded" message.

Another situation in which this often occurs is by right-clicking the Random button at https://scholia.toolforge.org/author/ multiple times in quick succession.

jdcaballerov:
@Daniel-Mietchen I think this PR https://github.com/WDscholia/scholia/pull/1652 will help a lot with this issue

Daniel-Mietchen:
I like the simplicity but have my doubts how far this will move things forward, since (i) lazy load needs to be enabled, (ii) it does not address the issue of multiple WDQS iframes loading in parallel, (iii) it does not catch the error.

I just tried this under Firefox with lazy load enabled and still got "Rate limit exceeded" errors in the scenarios listed above. I did not see any such errors for loading just a single page though, and this is indeed progress.

PR #883, opened 2019-10-25
Adds a chemistry/missing page for curation

egonw:
- with column on the `/chemical/` frontpage
- adds a few more example chemical structures 

I expect more patches later for more physchem properties from @lahire.

### DIFF
--- a/scholia/app/templates/chemicals_missing.html
+++ b/scholia/app/templates/chemicals_missing.html
@@ -9,7 +9,8 @@
 SELECT ?wikis ?compound ?compoundLabel ?formula WITH {
   SELECT ?wikis ?compound WHERE {
     ?compound wdt:P31 wd:Q11173 ;
-              wikibase:sitelinks ?wikis .
+              wikibase:sitelinks ?wikis . hint:Prior hint:rangeSafe true .
+    FILTER(?wikis >= 40)
     FILTER(NOT EXISTS {?compound wdt:P2102 []})
   } ORDER BY DESC(?wikis)
   LIMIT 100
@@ -24,7 +25,8 @@ SELECT ?wikis ?compound ?compoundLabel ?formula WITH {
 SELECT ?wikis ?compound ?compoundLabel ?formula WITH {
   SELECT ?wikis ?compound WHERE {
     ?compound wdt:P31 wd:Q11173 ;
-              wikibase:sitelinks ?wikis .
+              wikibase:sitelinks ?wikis . hint:Prior hint:rangeSafe true .
+    FILTER(?wikis >= 40)
     FILTER(NOT EXISTS {?compound wdt:P2101 []})
   } ORDER BY DESC(?wikis)
   LIMIT 100
@@ -39,7 +41,8 @@ SELECT ?wikis ?compound ?compoundLabel ?formula WITH {
 SELECT ?wikis ?compound ?compoundLabel ?formula WITH {
   SELECT ?wikis ?compound WHERE {
     ?compound wdt:P31 wd:Q11173 ;
-              wikibase:sitelinks ?wikis .
+              wikibase:sitelinks ?wikis . hint:Prior hint:rangeSafe true .
+    FILTER(?wikis >= 30)
     FILTER(NOT EXISTS {?compound wdt:P2067 []})
   } ORDER BY DESC(?wikis)
   LIMIT 100
@@ -54,7 +57,8 @@ SELECT ?wikis ?compound ?compoundLabel ?formula WITH {
 SELECT ?wikis ?compound ?compoundLabel ?formula WITH {
   SELECT ?wikis ?compound WHERE {
     ?compound wdt:P31 wd:Q11173 ;
-              wikibase:sitelinks ?wikis .
+              wikibase:sitelinks ?wikis . hint:Prior hint:rangeSafe true .
+    FILTER(?wikis >= 40)
     FILTER(NOT EXISTS {?compound wdt:P2177 []})
   } ORDER BY DESC(?wikis)
   LIMIT 100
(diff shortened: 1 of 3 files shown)


egonw:
Implements https://github.com/fnielsen/scholia/issues/546

fnielsen:
1. Headline level should be h2. May be a upper headline as h1 with "Missing information for chemical" or something. There is also a bit too much whitespace 
2. "sparqlToDataTable(missingDensitySparql, "#missing-density");" has too much whitespace. A couple or lines down also have too much whitespace.
3. Should the name of the template be different, e.g., chemical_empty_missing. The aspect_missing filenames are usually the ones pertaining to a specific Q-item, e.g., /scholia/chemical/Q123/missing. 
4. Perhaps we should be consistent with the case for the named subqueries. I hvae tended to use lowercase. I see that https://wiki.blazegraph.com/wiki/index.php/NamedSubquery uses lowercase+camelCase

Daniel-Mietchen:
Some more observations:
- the current queries do not seem to prioritize curation in any way (e.g. via sitelinks) but that would probably be helpful
- each of these tables is an opportunity to teach about the underlying concept, so "Missing Boiling Points" could link to [boiling point (Q1003183)](https://www.wikidata.org/wiki/Q1003183) and perhaps provide some [usage examples](https://www.wikidata.org/wiki/Property:P2102#P1855)
- some generic help link regarding how to handle units may also be useful
- typo: "Enthropy"

egonw:
@fnielsen, I suggest we make your point 3 and 4 issues, and do that later. (Done for point 3 already).

@Daniel-Mietchen, I like the idea for using sitelinks. I'm on it.  Also +1 on the documentation/guidance. Excellent point.

egonw:
hi @Daniel-Mietchen, @fnielsen, I tried to split nice-to-have from must-have-before-merge. Please have a look at the new patches, and let me know if there are further must-have-before-merge's.

scholia/app/templates/chemicals_missing.html
```diff
@@ -0,0 +1,309 @@
...
+  <li>use references: <a href="https://www.wikidata.org/wiki/Help:Sources">Help:Sources</a></li>
+</ul>
+
+The following links prioritize chemical compounds that are listed in most Wikis
+(Wikipedias, WikiCommons, etc). To go to the Wikidata page for each item, click the name
+below and then on the shown Scholia page for that compound the Q-identifier link at the top.
+
+<h2 id="h1">Missing Boiling Points</h2>
+
+The <a href="https://www.wikidata.org/wiki/Property:P2102">boiling point</a> is the temperature above
```
Daniel-Mietchen: That's not the definition of boiling point.

scholia/app/templates/chemicals_missing.html
```diff
@@ -0,0 +1,309 @@
...
+which the compound is liquid. Typical units given include Kelvin and degree Celsius. Because it
```
Daniel-Mietchen: Opportunity to link to "temperature" or "Kelvin" and "Celsius".

scholia/app/templates/chemicals_missing.html
```diff
@@ -0,0 +1,309 @@
...
+is depends on the pressure, this is often also given, as shown in
```
Daniel-Mietchen: Typo in "it is depends".

scholia/app/templates/chemicals_missing.html
```diff
@@ -0,0 +1,309 @@
...
+<table class="table table-hover" id="missing-pKa"></table>
+
+<h2 id="h1">Missing Combustion Enthalpy</h2>
+
+<table class="table table-hover" id="missing-combustion-enthalpy"></table>
+
+<h2 id="h1">Missing IDLH</h2>
+
+The <a href="https://www.wikidata.org/wiki/Property:P2129">Immediately Dangerous to Life and Health value</a>
+(IDLH) is an indication provided by the U.S.A. CDC about the danger of the compound.
```
Daniel-Mietchen: Opportunity to link to a CDC profile, perhaps https://tools.wmflabs.org/scholia/organization/Q583725 .

scholia/app/templates/chemicals_missing.html
```diff
@@ -0,0 +1,309 @@
...
+     sparqlToDataTable(missingCombustionEnthalpy, "#missing-combustion-enthalpy");
+     sparqlToDataTable(missingIDLH, "#missing-idlh");
+ });
+</script>
+ 
+{% endblock %}
+
+{% block page_content %}
+
+<h1>Missing physicalchemical properties for chemicals</h1>
```
Daniel-Mietchen: physicochemical

Daniel-Mietchen:
I left a few more comments in the code &mdash; nothing serious, and easy to fix. I also tried to load the missing page multiple times, and there was almost always some table that did not load, presumably due to the rate limit, as per #302 .

scholia/app/templates/chemicals_missing.html
egonw: Fixed.

scholia/app/templates/chemicals_missing.html
egonw: Linked to Kelvin and Celsius (and linked literature to both of them).

scholia/app/templates/chemicals_missing.html
egonw: Fixed.

scholia/app/templates/chemicals_missing.html
egonw: Added

scholia/app/templates/chemicals_missing.html
egonw: Fixed.

egonw:
Regarding the rate limit, I note too that they queries have become quite slow. This may have to do that Wikidata now has >1M compounds, instead of the 180k when these queries were made.

egonw:
It is the `wikibase:sitelinks` that makes it slow, it seems: https://phabricator.wikimedia.org/T175380

egonw:
Okay, the version now is a lot faster, BUT also returns less data. Particularly for more common properties this may be a problem, as it now only looks at compounds with at least 40 sitelinks (30 for molecular mass).

Daniel-Mietchen:
I think this is good to go now, even though there is also room for improvements, e.g. I think we should look into integrating with TABernacle for making the actual edits. Not sure it can do qualifiers though.

egonw:
`tox -e flake8` passed.

PR #909, opened 2019-11-03
code style fixes only

AlexanderPico:
Ok. Starting with a clean, rebased master, this patch only contains whitespace fixes, following the example by @fnielsen. There are no new features in this PR; just setting a clean background for this template.


AlexanderPico:
Process question @fnielsen: I see a prompt here to `delete branch` following the successful merge. Is that something I should do? Someone else? Or do you like keeping all patch branches?

fnielsen:
Hmmm... Good question. I see some discussion here: https://stackoverflow.com/questions/5330145/when-to-delete-branches-in-git

Maybe it is a good idea. I have a lot of branches (100) on my local computer and it clutters.

PR #915, opened 2019-11-12
adds description to pathway template

AlexanderPico:
- Makes use of extensible query for simple values to minimize number of queries
- Uses `div id` pattern
- Display of description matches that of other scholia templates that pull from Wikipedia (thus the `<br />`
- We are responsible for maintaining the wikidata content feeding this new description field and will keep them in sync if future changes are made

Example item with description data: https://tools.wmflabs.org/scholia/pathway/Q66104607

### DIFF
--- a/scholia/app/templates/pathway.html
+++ b/scholia/app/templates/pathway.html
@@ -92,12 +92,12 @@ ORDER BY DESC(?citations) DESC(?date)
 
 // Extensible query for muliple, optional values
 const optionalDataValuesSparql =`
-SELECT ?wpid ?organism ?organismLabel ?description
+SELECT ?wpid ?organism ?organismLabel ?pathwayDescription
 WHERE {
+  VALUES ?pathway { wd:{{q}} }
   OPTIONAL {
     wd:{{ q }} wdt:P2410 ?wpid ;
-    wdt:P703 ?organism ;
-    schema:description ?description .
+    wdt:P703 ?organism .
   }
   SERVICE wikibase:label { bd:serviceParam wikibase:language "[AUTO_LANGUAGE],en,da,de,es,fr,it,jp,nl,no,ru,sv,zh" . }
 }
@@ -121,8 +121,8 @@ $(document).ready(function() {
   $.getJSON(sparqlURL, function(response) {
     const simpleData = sparqlDataToSimpleData(response);
     const dataValues = simpleData.data[0];
-    if ("description" in dataValues) {
-      $("#description").text(dataValues.description);
+    if ("pathwayDescription" in dataValues) {
+      $("#description").text(dataValues.pathwayDescription);
     }
     if ("organism" in dataValues) {
       $("#Organism").after('<a href="'+dataValues.organism+'">'+dataValues.organismLabel+'</a>'); 
--- a/scholia/app/templates/pathway.html
+++ b/scholia/app/templates/pathway.html
@@ -92,11 +92,12 @@ ORDER BY DESC(?citations) DESC(?date)
 
 // Extensible query for muliple, optional values
 const optionalDataValuesSparql =`
-SELECT ?wpid ?organism ?organismLabel
+SELECT ?wpid ?organism ?organismLabel ?description
 WHERE {
   OPTIONAL {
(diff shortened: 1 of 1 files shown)


AlexanderPico:
@egonw @fnielsen Next PR, based on lastest master including recent code style refactor. Let me know if you have questions.

scholia/app/templates/pathway.html
```diff
@@ -92,11 +92,12 @@
 
 // Extensible query for muliple, optional values
 const optionalDataValuesSparql =`
-SELECT ?wpid ?organism ?organismLabel
+SELECT ?wpid ?organism ?organismLabel ?description
 WHERE {
   OPTIONAL {
     wd:{{ q }} wdt:P2410 ?wpid ;
     wdt:P703 ?organism ;
+    schema:description ?description .
```
fnielsen: This will result in multiple values for each language AFAIU.

fnielsen: What about `VALUES ?pathway { wd:{{ q }} }` and `?pathwayDescription`?

AlexanderPico: Nice! That works better. Change pushed. Should be part of this PR now.

AlexanderPico:
@fnielsen Any other suggests or issues on this one?

scholia/app/templates/pathway.html
```diff
@@ -120,6 +121,9 @@
   $.getJSON(sparqlURL, function(response) {
     const simpleData = sparqlDataToSimpleData(response);
     const dataValues = simpleData.data[0];
+    if ("description" in dataValues) {
+      $("#description").text(dataValues.description);
+    }
     if ("organism" in dataValues) {
       $("#Organism").after('<a href="'+dataValues.organism+'">'+dataValues.organismLabel+'</a>');
```
fnielsen: Do we need an escape of dataValues.organismLabel?

Sorry for the delay in my response.

AlexanderPico: I probably just need coffee... but I'm not sure what you mean by escape and why we would need one here?  And you say, "dataValues.organismLabel", but I'm assuming you mean, "dataValues.description."

Do you mean, we need an "else" case? The "else" is to do nothing, i.e., not to add any text to the `description` div. It will simply be blank if no description is available (like it is currently).

fnielsen: I mean in the unlikely case that dataValues.organismLabel contains, e.g., "<script..." the the "<" should be escaped. I suppose "dataValues.organism" cannot be other than Q\d+ so that is not a problem. Possibly Q4115189 can be used.

AlexanderPico: Ok. Happy to do that. I've opened an issue to do this in another PR, however, since this one is just about adding the description handling code. The line you want changed is not part of this PR.

https://github.com/wikipathways/scholia/issues/4

AlexanderPico: @fnielsen Looks like I should add the escaping of "<" to this utils function and then use it for dataValues.organismLabel: https://github.com/fnielsen/scholia/blob/master/scholia/utils.py#L7. Right?

And I can use sanitize_q on dataValues.organism.

Again, I'll do this in a separate PR. Anything else for this PR on description handling?

fnielsen: I wonder if `.text(dataValues.organismLabel)` can be used? This is already done above. I suppose the Q-identifier would need no sanitization.

AlexanderPico: This PR is a few lines about adding descriptions to the template. Your questions appear to be about how the organism info could be improved. That code was merged in a prior PR and is already live. 

I've started a new issue (https://github.com/wikipathways/scholia/issues/4) to work on those suggestions, but don't want to prepare the next PR until this one is clear. 

Any other questions or suggestions for this PR?

fnielsen:
I do not understand why the link under "Homo sapiens" is going to Wikidata rather than to Scholia. Was that intended?

AlexanderPico:
Ah, we probably didn't realize that Scholia had a Taxon template at the time. Good catch. I'll add that to the list of fixes for organism handling: https://github.com/wikipathways/scholia/issues/4

Issue #959, opened 2019-12-02
Add favicon


Daniel-Mietchen:
Perhaps using our logo as a starting point: https://commons.wikimedia.org/wiki/File:Scholia_logo.svg .

fnielsen:
Are we able to do that? We do not have the top level of tools.wmflabs.org?

Daniel-Mietchen:
Judging on the basis of things like https://tools.wmflabs.org/reasonator/ , there should be a way.

carlinmack:
I think I can get this done wednesday morning, I'm quite familiar with favicons

carlinmack:
Is there a urls.py file for in scholia? or how are pages served, I want to create a redirect for /favicon.ico to the static path

fnielsen:
I am wondering whether HTML base tag could/should be used (too)

Daniel-Mietchen:
Users are beginning to notice: https://twitter.com/GerardMeijssen/status/1202466066175266816

fnielsen:
The favicon is now running.

I had difficulty reviewing the code as I am not familiar with favicon. I thought it was just a question of putting something in the root directory, but I see it is somewhat more elaborate.

carlinmack:
[Need to add a few more redirects](https://realfavicongenerator.net/favicon_checker?protocol=https&site=tools.wmflabs.org%2Fscholia%2F#.Xeh5DxCYWNE) but for 99% of real use cases it's fine :)

carlinmack:
Actually reading what that says, I don't know if we can fully resolve as we aren't serving our site from the / root domain. Therefore I don't think there is much that can be done but that's not a bad thing :)

Daniel-Mietchen:
Looks good enough for the moment. Close?

carlinmack:
👍

PR #973, opened 2019-12-04
Add favicon files and html to header

carlinmack:
Closing #959 

image

image

image


scholia/app/views.py
```diff
@@ -1857,3 +1856,8 @@ def show_about():
 
     """
     return render_template('about.html')
+
+
+@main.route('/favicon.ico')
+def show_test():
```
fnielsen: I am not sure why this is called show_test and not, e.g., show_favicon.

carlinmack: good catch I called it show test as I was testing how flask routing works, I'll change it

Issue #985, opened 2019-12-08
In missing topics panel on missing page for authors, sort by citation count


Daniel-Mietchen:
example:
https://tools.wmflabs.org/scholia/author/Q97270/missing

PR #986, opened 2019-12-08
sorted missing topics panel on missing author page

Daniel-Mietchen:
I added sorting to the results table, so as to help prioritize curation.


Daniel-Mietchen:
Fixes #985 .

Issue #786, opened 2019-07-17
Continous integration – run tests automatically


rasmuserik:
It would be nice to run tests automatically in pull-requests etc.

I propose setting up Travis-CI

nunogit:
@Daniel-Mietchen I am working on this and made a Travis test in my fork, I see that the checkbox for " make .travis.yml that runs the tox test" is ticked however I don't find the .travis.yml
Is this already done?  Than you.

Daniel-Mietchen:
It [was](https://github.com/rasmuserik/scholia/commit/8502bbe920392c99e1af72e019b9a9f45c7f5f55) in the pull requests by @rasmuserik that are linked above and for some reason did not make it into the master branch back then.

fnielsen:
There is a test at https://github.com/WDscholia/scholia/blob/master/.github/workflows/main.yml 

I am wondering what else there is needed beyond this kind of test?

Daniel-Mietchen:
Some other things that come to mind:
- testing for proper formatting, e.g. via [black](https://pypi.org/project/black/)
- autogenerate documentation, e.g. via [Sphinx](https://www.sphinx-doc.org/en/master/index.html)

fnielsen:
We already test for proper formatting with pydocstyle. Is black better?

PR #1011, opened 2020-01-06
Aims at solving #786

nunogit:
This pull request aims at solving Continuous integration – run tests automatically, issue #786

Tox is now running successfully (tests).

Travis is not yet running successfully. https://www.travis-ci.com/github/scholia
An email was sent to Travis to activate debug mode and understand why Travis is not working properly yet.
You can find Travis reports here: https://travis-ci.org/nunogit/scholia

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: python
+python:
+  - "3.4"
+  - "3.5"
+  - "3.6"      # current default Python on Travis CI
+  - "3.7"
+  - "3.8"
+#  - "3.8-dev"  # 3.8 development branch
+#  - "nightly"  # nightly build
+# command to install dependencies
+install:
+  - pip install -r requirements.txt
+# command to run tests
+script:
+  - pytest
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ python:
 # command to install dependencies
 install:
   - pip install -r requirements.txt
+  - pip install tox
 # command to run tests
 script:
-  - pytest
+  - tox
+#  - pytest
--- a/runserver.py
+++ b/runserver.py
@@ -7,4 +7,4 @@ app = create_app(
 app.config['APPLICATION_ROOT'] = '/'
 
 if __name__ == '__main__':
-    app.run(debug=True, port=8100)
+    app.run(debug=True, host='0.0.0.0', port=8100)
--- a/runserver.py
+++ b/runserver.py
@@ -7,4 +7,4 @@ app = create_app(
 app.config['APPLICATION_ROOT'] = '/'
 
 if __name__ == '__main__':
-    app.run(debug=True, host='0.0.0.0', port=8100)
+    app.run(debug=True, host='0.0.0.0',  port=8100)
--- a/scholia/network.py
+++ b/scholia/network.py
@@ -34,7 +34,8 @@ def write_pajek_from_sparql(filename, sparql):
     column1 = 'item1'
     column2 = 'item2'
 
-    url = 'https://query.wikidata.org/sparql'
+    #url = 'https://query.wikidata.org/sparql'
+    url = 'http://127.0.0.1:8080/sparql'
     params = {'query': sparql, 'format': 'json'}
     response = requests.get(url, params=params)
     data = response.json()['results']['bindings']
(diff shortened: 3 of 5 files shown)


fnielsen:
I think Travis wants access to private repositories (I wonder why that is necessary?) and I would like to move one before given it access.

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  - "3.4"
+  - "3.5"
+  - "3.6"      # current default Python on Travis CI
+  - "3.7"
+  - "3.8"
+#  - "3.8-dev"  # 3.8 development branch
+#  - "nightly"  # nightly build
+# command to install dependencies
+install:
+  - pip install -r requirements.txt
```
cthoyt: you shouldn't have to install the requiements.txt file since tox will install the whole scholia package using `pip install .`.

tox.ini
```diff
@@ -1,16 +1,16 @@
...
-    -rrequirements.txt
+#[testenv:py27]
+#commands = 
+#    python -m pytest --doctest-modules scholia
+#    python -m pytest tests
+#deps=
+#    pytest
+#    -rrequirements.txt
 
-[testenv:py35]
```
cthoyt: if you just the magical header `[testenv]` then it uses whatever version of python on which tox is installed

tox.ini
```diff
@@ -1,16 +1,16 @@
 [tox]
-envlist = flake8, pydocstyle, py27, py35
+envlist = py36
```
cthoyt: if you use the magical environment `py` then it will default to whatever version of python is available.

nunogit: Thank you for your input. My question then is - should we do that, or make sure it works well with a specific version(s) so that we can advise that to whoever wants to install?

scholia/network.py
```diff
@@ -34,7 +34,8 @@ def write_pajek_from_sparql(filename, sparql):
     column1 = 'item1'
     column2 = 'item2'
 
-    url = 'https://query.wikidata.org/sparql'
+    #url = 'https://query.wikidata.org/sparql'
+    url = 'http://127.0.0.1:8080/sparql'
```
fnielsen: Why are these host changed?

tox.ini
cthoyt: Both are possible - Travis can test on several versions of python, and it looks like you've already put them in the `.travis.yml`. Further, you can see from the last build log at https://travis-ci.org/nunogit/scholia/jobs/633355631 that there's a mismatch between the version of python available when specifying it this way, and the version tox wants to run.

However, I would suggest removing python34. It's really old and most libraries aren't supporting it anymore (see https://travis-ci.org/nunogit/scholia/jobs/633355625)

cthoyt:
It would also be cool to put a Travis badge on the README.rst so readers can quickly find the builds. You can use
```rst
.. image:: https://travis-ci.org/nunogit/scholia.svg?branch=master
    :target: https://travis-ci.org/nunogit/scholia
```

But `travis-ci.org` should probably be switched to `travis-ci.com` based on where @fnielsen activates it, and switching `nunogit/scholia` to `fnielsen/scholia`

runserver.py
```diff
@@ -7,4 +7,4 @@
 app.config['APPLICATION_ROOT'] = '/'
 
 if __name__ == '__main__':
-    app.run(debug=True, port=8100)
+    app.run(debug=True, host='0.0.0.0', port=8100)
```
cthoyt: In docker it would be much preferred to use a proper server and not the test one provided by flask/Werkzeug. Maybe in a different PR?

nunogit: @cthoyt, for now, this is a prototype to test some other aspects (not really meant for production).

In any case, would you care to give an example just to understand better what you mean?

tox.ini
nunogit: Ok, now I understand what you mean. Agreed!

egonw:
@fnielsen, what is the next step for this PR?

runserver.py
cthoyt: @nunogit I actually wrote up a bit on this the other day, might be interesting for you: https://cthoyt.com/2021/01/11/click-and-flask.html#using-gunicorn

egonw:
@fnielsen, are you okay to close this one, as outdated?

Daniel-Mietchen:
Closing - too old now, too out of sync with the code base.

PR #1018, opened 2020-01-13
Closing #608

Daniel-Mietchen:
Modified the query for "Author name strings to be resolved" on the /missing page for authors, such that it includes strings that occur in "stated as" qualifiers for P50 (author) statements involving this entity on publicaiton items.


Daniel-Mietchen:
I checked this once more, and it looks fine.

PR #1037, opened 2020-02-06
Creating a configuration file in Scholia

nunogit:
Creating a configuration file in Scholia to make the data sources (SPARQL EndPoint) configurable.
This is very important to allow caching servers in an easy and transparent way. An added benefit is the fact that we can now support new data sources (SPARQL end-points) in an easy way.

I would appreciate comments, especially on the variable names of the config file and in the way they were set (3 servers).


scholia/config.py
```diff
@@ -0,0 +1,46 @@
...
+CONFIG_FILENAMES = [
+    'scholia.cfg',
+    '~/etc/scholia.cfg',
+    '~/scholia.cfg']
+
+DEFAULTS = """
+[servers]
+SPARQLEndPointURLbd = https://query.wikidata.org/bigdata/namespace/wdq/sparql
+SPARQLEndPointURL = https://query.wikidata.org/sparql
+SPAREndPointEmbed = https://query.wikidata.org/
```
fnielsen: Shouldn't this be SPARQLEndpointEmbed or QueryServiceURL

scholia/config.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+CONFIG_FILENAMES = [
+    'scholia.cfg',
+    '~/etc/scholia.cfg',
+    '~/scholia.cfg']
+
+DEFAULTS = """
+[servers]
+SPARQLEndPointURLbd = https://query.wikidata.org/bigdata/namespace/wdq/sparql
```
fnielsen: Is this variable necessary?

scholia/network.py
```diff
@@ -8,6 +8,7 @@
 from collections import OrderedDict
 
 import requests
+from scholia.config import config
```
fnielsen: Shouldn't this  be `from .config import config`

scholia/api.py
```diff
@@ -156,7 +156,7 @@ def wb_get_entities(qs):
         'User-Agent': 'Scholia',
     }
     response_data = requests.get(
-        'https://www.wikidata.org/w/api.php',
+        config.get('servers', 'webserviceURL'),
```
fnielsen: This variable is missing from the defaults. Shouldn't it be "WebserviceURL" if the CamelCase is followed?

scholia/app/templates/base.html
```diff
@@ -104,7 +106,7 @@
 
      });
 
-     var endpointUrl = 'https://query.wikidata.org/sparql';
+     var endpointUrl = '{{ webserviceURL }}sparql';
```
fnielsen: This seems to be wrong.

scholia/config.py
nunogit: I have pushed a new batch of improvements. The naming wasn't consistent.

scholia/config.py
nunogit: That's a good question. 
This URL is used in a couple of places in scholia. 
If it makes sense to use a bigdata separate endpoint than it makes sense to make it into a configuration also.

Note:
This could also be done differently, by having a single variable for the SPARQL endpoint and appending the suffix in the code, depending on if you want to access bigdata or just the normal endpoint. My choice keeps more flexibility since we can use completely different URLs for bigdata and normal queries.

scholia/api.py
nunogit: This was one of the things to fix

scholia/app/templates/base.html
nunogit: Fixed. 
I tested everything before committing. Everything seemed to work.
These URLs are spread all around. It's easy to miss. Thanks

scholia/config.py
fnielsen: I think that `https://query.wikidata.org/bigdata/namespace/wdq/sparql` is just an old endpoint and we should have changed that long ago. Are there really difference between that and /sparql endpoint?

nunogit: I have no idea. I thought that had some special purpose.
I will remove that configuration and set the URL to the main one in scholia if you agree.

fnielsen: I have now asked https://www.wikidata.org/w/index.php?title=Wikidata_talk:SPARQL_query_service&diff=1111375482&oldid=1053274607

scholia/network.py
fnielsen: I think this should be a relative import, i.e., `from .config import config` instead of `from scholia import config`

scholia/config.py
nunogit: @fnielsen do you think you can accept this pull request and, depending on the answer, do further adaptations?
I have a couple of other pull requests to commit and request. I don't want to do that before closing this one though.

fnielsen:
The branch should have had a more telling name

runserver.py
```diff
@@ -7,4 +7,4 @@
 app.config['APPLICATION_ROOT'] = '/'
 
 if __name__ == '__main__':
-    app.run(debug=True, host=0.0.0.0,  port=8100)
+    app.run(debug=True, host='0.0.0.0',  port=8100)
```
fnielsen: Isn't a default 0.0.0.0 development IP a security risk?

fnielsen: Why is it necessary with the host IP there?

nunogit: The default is 127.0.0.1 (localhost). The need to add host 0.0.0.0 is to bind to all interfaces.
This is relevant for Docker, otherwise, it doesn't work.
We can also externalize this to the config file. It would make sense.
What do you think @fnielsen ?

fnielsen: Default on localhost and then perhaps configurable.

fnielsen:
* Run Tox.
* Python variables should be lowercase.
* I think it should be Endpoint instead of EndPoint. 

``` 
SPARQLEndpointURL sparql_endpoint_url
SPARQLEmbedURL sparql_embed_url
WebserviceURL webservice_url
``
It is unclear to me whether there is any particular format for the INI file. When I look at 
https://docs.python.org/3/library/configparser.html is seems

File: runserver.py (hunk 6 above)
Comment (nunogit) [reply runserver.py]: Done as suggested.

**Comment by fnielsen** [comment]
What is `report_error`?

**Comment by fnielsen** [comment]
flake8 tels about many errors.

**Comment by nunogit** [comment]
> What is `report_error`?

@fnielsen  shouldn't be there. It's something I  am working on to get feedback from users whenever they find an error or if by their initiative they wish to give feedback. 
Issues #462 #1026; Ultimately contribute to a dynamic list of pages to test.

> flake8 tels about many errors.

Most of the style errors I am getting are from behind. I can fix those, but it would be better to handle that in a new pull request/issue? This task is already very wide. What do you think?

**Comment by fnielsen** [comment]
You should run `tox` before making a PR and make sure there is no errors.

**Comment by nunogit** [comment]
@fnielsen fixed most errors. Left long lines and a couple of remarks about variables (not sure if there is a plan for them)

**Comment by fnielsen** [comment]
Why is there a 
```
flask-sessionstore
flask-session-captcha
``` 
in the requirements.txt

**Comment by fnielsen** [comment]
`scholia/api.py` line 160 seems to be missing the configuration

**Comment by fnielsen** [comment]
`/scholia/app/static/scholiaConfig.js`: Why is the Javascript constant not setup via the configuration from Python?

**Comment by fnielsen** [comment]
In `/scholia/app/templates/author.html`  "bigdata/namespace/wdq/" is not necessary AFAIK.

**Comment by fnielsen** [comment]
Why are there change in  `/scholia/app/templates/author_empty.html` unrelated to the configuration?

**Comment by fnielsen** [comment]
There are some new commit that are not added in `/scholia/app/templates/base.html`

**Comment by fnielsen** [comment]
In `/scholia/app/templates/base.html` there are some letovers from report-error.

**Comment by fnielsen** [comment]
`/scholia/app/templates/event.html`  line 308 has a missing commit.

**Comment by fnielsen** [comment]
`/scholia/app/templates/event_series.html` has a missing commit.

**Comment by fnielsen** [comment]
`/scholia/app/templates/pathway.html` line 116 has unnecessary part in the  sparqlURL.

**Comment by fnielsen** [comment]
There is a report_error.html

**Comment by fnielsen** [comment]
Missing commit in `/scholia/app/templates/venue_empty.html`

**Comment by fnielsen** [comment]
`from requests.auth import HTTPBasicAuth`?

**Comment by fnielsen** [comment]
Why is `current_app` renamed?

**Comment by fnielsen** [comment]
Why do we need. 

```
from flask_sessionstore import Session
from flask_session_captcha import FlaskSessionCaptcha
```

fnielsen:
The is a missing commit in `/scholia/arxiv.py`

nunogit:
I just finished the changes as discussed in our previous meeting.

egonw:
@fnielsen, what is the next action for this PR?

egonw:
@fnielsen, are you okay to close this one, as outdated?

Daniel-Mietchen:
Closing - too old now, too out of sync with the code base.

PR #1044, opened 2020-02-17
Give the full InChI (closes #948)

### LINKED ISSUE #948: InChI not shown in full
It visualizes the InChI as given in Wikidata, and lacks the `InChI=` prefix.

egonw:


Daniel-Mietchen:
Works fine. Does/ should something similar apply to InChIKey?

egonw:
> Does/ should something similar apply to InChIKey?

I need to check. Normally it doesn't come with a prefix, but I have seen that recently. I would suggest not to use it.

egonw:
I checked. Formally, the InChIKey is with a `InChIKey=` prefix. But I still don't want to add that. Common practice there is really without.

fnielsen:
I do not understand the headers "IDpred" and "Finalid". Could we give the some names that I can understand?

Issue #1045, opened 2020-02-17
add Twitter handle and hashtag to base.html


egonw:

egonw:
See https://github.com/fnielsen/scholia/pull/1047

fnielsen:
Do we have an example event with a Twitter handle and a Twitter hashtag?

egonw:
- https://tools.wmflabs.org/scholia/event-series/Q77165314
- https://tools.wmflabs.org/scholia/event/Q77165107

fnielsen:
And https://tools.wmflabs.org/scholia/event/Q64682140

PR #1047, opened 2020-02-18
adds Twitter accounts and the hashtag (implements #1045)

egonw:
- to pages that already showed some info
- and details are now shown for events and event series too
- Twitter account 
- hashtag, which points to the wmflabs hashtags-hub service

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -56,7 +56,12 @@
 	     $( '#details' ).append( '<a href="' + fullTextUrl + '">' + fullTextUrl + '</a>. ');
 	 }
 	 catch(e) {}
-	 
+
+	 try {
+	     var twitterAccount = item.claims.P2002[0].mainsnak.datavalue.value;
+	     $( '#details' ).append( '<a href="https://twitter.com/' + twitterAccount + '">@' + twitterAccount + '</a> ');
+	 }
+	 catch(e) {}
 
 	 /* English Wikipedia */
 	 if ('enwiki' in item.sitelinks) {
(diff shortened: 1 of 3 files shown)


Daniel-Mietchen:
The current implementation does not indicate that the handle is for Twitter - perhaps add a [Twitter logo](https://commons.wikimedia.org/wiki/Category:Twitter_logos) (perhaps [this t](https://commons.wikimedia.org/wiki/File:Twitter_logo_initial.svg) or [this bird](https://commons.wikimedia.org/wiki/File:Twitter-Logo.svg)).

egonw:
Logo added:

image

Daniel-Mietchen:
Looks good to me now, also for hashtags.

scholia/app/templates/base.html
```diff
@@ -56,7 +56,12 @@
 	     $( '#details' ).append( '<a href="' + fullTextUrl + '">' + fullTextUrl + '</a>. ');
 	 }
 	 catch(e) {}
-	 
+
+	 try {
+	     var twitterAccount = item.claims.P2002[0].mainsnak.datavalue.value;
+	     $( '#details' ).append( '<a href="https://twitter.com/' + twitterAccount + '">@' + twitterAccount + '</a> ');
```
fnielsen: I wonder if we might want a escape here, - just in case? I see that the above fullTextUrl does not have one...

egonw: Escape of what? I think the Twitter account name is meant to give valid URLs.

fnielsen:
I have begone to work a bit on the code here.

Issue #1043, opened 2020-02-14
In work aspect, add most frequently cited authors


Daniel-Mietchen:
I.e. rank authors by number of works cited from this work that were (co-)authored by them.

PR #1057, opened 2020-02-22
Add most cited authors (solves issue #1043)

awakenting:
Hi,

I wanted to try out some SPARQL queries and I think I managed to implement the request in [issue 1043](https://github.com/fnielsen/scholia/issues/1043) with this pull request.

I also tested it locally and looked at a few example works. The tests with tox failed the same way as they did right after cloning, so I thought that's good enough.

### DIFF
--- a/scholia/app/templates/work.html
+++ b/scholia/app/templates/work.html
@@ -134,29 +134,26 @@ WHERE {
 ORDER BY DESC(?itemLabel)
 `;
 
-citedWorksAuthorsSparql = `
+ citedWorksAuthorsSparql = `
 #defaultView:Table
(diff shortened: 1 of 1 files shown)


Daniel-Mietchen:
Thanks, @awakenting !
This looks good for the most part. I suggest to add a DISTINCT in order to account for cases where authors have more than one ORCID, e.g. https://www.wikidata.org/wiki/Q573389 , which appears in the results for https://www.wikidata.org/wiki/Q24302302 .

fnielsen:
I am wondering how relevant the ORCID is. I do rarely use it.

Issue #461, opened 2018-08-21
"Recently published works on the chemical" times out for fatty acids in chemical-class aspect


Daniel-Mietchen:
https://tools.wmflabs.org/scholia/chemical-class/Q61476

This probably means that the inner query needs a LIMIT too.

egonw:
It seems to work for me know... introducing a LIMIT would cause to not all literature get counted, or?

Another though: was that trick already introduced in the query to first group by year?

fnielsen:
This gets the publication date in the first SELECT: 

```
SELECT ?date ?work ?workLabel ?type ?topics
WITH {
  SELECT (MAX(?dates) as ?datetime) ?work WHERE {
    {
      ?work wdt:P921 / (wdt:P361+ | wdt:P1269+ | (wdt:P31* / wdt:P279*) ) wd:Q61476 .
    } UNION {
      wd:Q61476 ?propp ?statement .
      ?statement a wikibase:BestRank ;
                 prov:wasDerivedFrom/pr:P248 ?work .
    }
    OPTIONAL { ?work wdt:P577 ?dates . }
  }
  GROUP BY ?work
  ORDER BY DESC(?dates)
  LIMIT 500
} AS %works
WITH {
  SELECT
    ?datetime
    ?work
    (GROUP_CONCAT(DISTINCT ?type_label; separator=", ") AS ?type)
    (GROUP_CONCAT(?topic_label; separator=" // ") AS ?topics)
  WHERE {
    INCLUDE %works
    ?work wdt:P921 ?topic . 
    OPTIONAL { ?work wdt:P31 ?type_ . ?type_ rdfs:label ?type_label . FILTER (LANG(?type_label) = 'en') }
    ?topic rdfs:label ?topic_label .  FILTER (lang(?topic_label) = 'en')
  }
  GROUP BY ?work ?datetime
} AS %result
WHERE {
  INCLUDE %result
  BIND(xsd:date(?datetime) AS ?date)
  SERVICE wikibase:label { bd:serviceParam wikibase:language "en,da,de,es,fr,jp,nl,no,ru,sv,zh". }
}
ORDER BY DESC(?date)
LIMIT 500
```

egonw:
Still works for me. Can this be closed?

Daniel-Mietchen:
Still does not work for me but is addressed by PR #1063 , so when that is deployed, we can close this ticket here too.

PR #1063, opened 2020-02-29
better literature for chemicals and fixed 'fatty acid' literature

egonw:
The problem here was that it also showed biological pathways. This is fixed by inverting the 'part of'. When debugging, I extended the query to record the 'via', which I found sufficiently helpful, that I kept that too.


egonw:
I actually ran into #461 and applied that query update there too, as well as the 'date' trick reported in that issue.

Daniel-Mietchen:
Looks good to me.

egonw:
`tox -e flake8` fails on the branch, but that's due to a fix in the master commit it was branched from. I manually/locally merged into master (on a test branch), and then `tox -e flake8` comes out green.

fnielsen:
I am wondering whether we have an example of a chemical where (all) the combination of "via" is there, - just to see that it works.

PR #1068, opened 2020-03-10
better text about author disambiguation on the missing page for authors

Daniel-Mietchen:
better text about author disambiguation on the missing page for authors


Daniel-Mietchen:
Took another look at it and think it's fine. Merging.

Issue #1084, opened 2020-03-24
In "Map of organizations associated with works about the topic", make colour scale more useful


Daniel-Mietchen:
e.g. 
- https://tools.wmflabs.org/scholia/topic/Q12156 (three layers, with little differentiation at the lower end)
  - solution: increase LIMIT
- https://tools.wmflabs.org/scholia/topic/Q84263196 (only one layer)
  - solution: add "1" as a separate layer and rearrange accordingly

<img width="1123" alt="Screen Shot 2020-03-24 at 00 52 24" src="https://user-images.githubusercontent.com/465923/77389654-1e44ae80-6d6a-11ea-8e0e-0d9b001a13ea.png">
<img width="1130" alt="Screen Shot 2020-03-24 at 00 53 02" src="https://user-images.githubusercontent.com/465923/77389659-2270cc00-6d6a-11ea-9867-1a78b67e23cc.png">

Daniel-Mietchen:
Closed by #1085

PR #1085, opened 2020-03-24
improved colour scale for topic orgs map

Daniel-Mietchen:
Initial attempt to address #1084


Daniel-Mietchen:
Looks better. Probably need to drop the "results" from "10 results". 

Screen Shot 2020-03-24 at 01 11 03
Screen Shot 2020-03-24 at 01 11 55

Daniel-Mietchen:
Would be great to arrange the colours such that the least occupied layer (i.e. highest numbers) are always displayed on top. May require changing the direction of the IF clause.

Daniel-Mietchen:
I played around with color rearrangement, but it does not seem to be possible to specify the order in which the layers are being stacked for display.

Daniel-Mietchen:
In terms of changing direction of the IF clause, that query is [here](https://www.wikidata.org/w/index.php?title=User:Daniel_Mietchen/Sandbox4URLshortening&oldid=1144061778).

Daniel-Mietchen:
Not sure what you mean by the fewer dots. If that refers to the two pics in https://github.com/fnielsen/scholia/pull/1085#issuecomment-603021008 , then they do not have to be compared against each other but against the two pics in https://github.com/fnielsen/scholia/issues/1084#issue-586691879 , which represent two different use cases behind the two changes in the patch (removing the LIMIT, adding 1 as another step).

Daniel-Mietchen:
Not sure about the explanation of the dot colour either, but I just added #1087 for that.

egonw:
> Not sure what you mean by the fewer dots. If that refers to the two pics in [#1085 (comment)](https://github.com/fnielsen/scholia/pull/1085#issuecomment-603021008) , then they do not have to be compared against each other

Oh, sorry, I thought they were before after :)

egonw:
> Not sure about the explanation of the dot colour either, but I just added #1087 for that.

Ah. awesome. @Daniel-Mietchen, you can actually push additional patches to existing pull requests. I will see if I can merge in #1087 into this branch.

Daniel-Mietchen:
I purposefully kept the explanation commit separate, since I am not sure that's what we need. But I take your merge as a yes and am merging all of it into master now.

PR #1090, opened 2020-03-28
Added NCBI gene and taxon IDs to the UI

egonw:
Adds two CLI methods:

* `python -m scholia.query ncbitaxon-to-q 694009`
* `python -m scholia.query ncbigene-to-q 5604`

Second, it upgrades the empty `gene` and `taxon` aspects with the column layout, featuring the two redirects, here for taxon:

image

**Important**: unlike with most other redirects, I decided here to use simply `ncbi`, but because ncbi has many identifiers, I use the aspect as context.

### DIFF
--- a/scholia/app/templates/gene_empty.html
+++ b/scholia/app/templates/gene_empty.html
@@ -45,7 +45,7 @@ Genes.
 		<p>If you know the identifier then Scholia can make a lookup based on the identifier:</p>
 
 		<dl>
-		    <dt><a href="ncbi/5604">gene/ncbi/5604</a></dt>
+		    <dt><a href="../ncbi-gene/5604">ncbi-gene/5604</a></dt>
 		    <dd>Information on MAP2K1 based on the NCBI (Entrez) gene identifier.</dd>
 		</dl>
 	    </div>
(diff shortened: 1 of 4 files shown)


egonw:
Fixed a small whitespace issue. 

`tox -e flake8` fails on the branch, but that's due to a fix in the master commit it was branched from. I manually/locally merged into master (on a test branch), and then `tox -e flake8` comes out green (after the fix).

fnielsen:
@egonw What is the reason to have multiple "directories", e.g., `/taxon/ncbi/<taxon>` in the route? Why not just `/ncbi/`?

egonw:
> @egonw What is the reason to have multiple "directories", e.g., `/taxon/ncbi/<taxon>` in the route? Why not just `/ncbi/`?

Because the NCBI database has identifiers for taxons, genes, proteins, transcripts, and a bunch more. And since they are merely numbers, you cannot see from the identifier (or match against Wikidata), which "type" it is. I have considered just `/ncbitaxon/` but found the multiple directories more elegant, personally.

fnielsen:
What about `ncbi-taxon` and `ncbi-gene` for both URL and command line? To me that seems more readable.

egonw:
I rebased the branch on the latest master.

egonw:
@fnielsen, I've updated the patch.

PR #1091, opened 2020-03-28
Better column names

egonw:
Addresses https://github.com/fnielsen/scholia/issues/1070

### DIFF
--- a/scholia/app/templates/chemical.html
+++ b/scholia/app/templates/chemical.html
@@ -9,11 +9,11 @@
  structuresSparql = `
 SELECT
   ?IDpred ?IDpredLabel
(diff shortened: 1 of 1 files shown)


egonw:
`tox -e flake8` fails on the branch, but that's due to a fix in the master commit it was branched from. I manually/locally merged into master (on a test branch), and then `tox -e flake8` comes out green.

fnielsen:
@egonw:

* Should "Structural identifier" be lowercased? 
* The idUrl in the structural table seems not to be used.
* I am trying to understand the `?idLit variable.` It is use in the result of a subquery, but not the upper query.

Issue #1101, opened 2020-04-06
On topic profiles, add country-level citation network


Daniel-Mietchen:
Ideally, the entries in the graph would link directly to the country/ topic profiles, e.g. as per
https://tools.wmflabs.org/scholia/country/Q668/topic/Q10737 .

Daniel-Mietchen:
[Here](https://commons.wikimedia.org/wiki/File:Country-level_citation_graph_for_Zika_virus_publications,_as_per_Wikidata_Query_Service_on_6_April_2020.png) is a draft query for such a country-level citation network, using Zika virus as an example:

[![Country-level_citation_graph_for_Zika_virus_publications](https://upload.wikimedia.org/wikipedia/commons/thumb/d/db/Country-level_citation_graph_for_Zika_virus_publications%2C_as_per_Wikidata_Query_Service_on_6_April_2020.png/686px-Country-level_citation_graph_for_Zika_virus_publications%2C_as_per_Wikidata_Query_Service_on_6_April_2020.png)](https://commons.wikimedia.org/wiki/File:Country-level_citation_graph_for_Zika_virus_publications,_as_per_Wikidata_Query_Service_on_6_April_2020.png)

Daniel-Mietchen:
On that basis, I played around with a suitable number of connections, and 30 to 50 seems to be useful for most cases.

egonw:
What happens for multi-country articles that cite another multi-country article? I can imagine these graphs to very quickly become a network blob.

Daniel-Mietchen:
@egonw In the query used in the PR at #1102 , there is a limit of the number of connections displayed, which I arbitrarily set to 42 after playing around with values between about 10 and about 100 for well-curated topics.

fnielsen:
This is an interesting graph. It must be expensive if there are many papers. Wikipedia and Machine learning and biophysics worked fast though.

fnielsen:
It is now running, e.g., at https://tools.wmflabs.org/scholia/topic/Q9035

PR #1102, opened 2020-04-06
add country-level citation graph per topic

Daniel-Mietchen:
First attempt at #1101.

Worked for all the about two dozen topics I tested, except for https://tools.wmflabs.org/scholia/topic/Q3450985 (neurodevelopmental disorders), which gave an empty response.


Daniel-Mietchen:
Another quirk: for microstructure, the flag for South Africa was just shown as the flag from the Apartheid period. This was easily [fixed](https://www.wikidata.org/w/index.php?title=Q160250&diff=prev&oldid=1151580302) but should be kept in mind.

Also, while tracking this down, I was reminded that we have [flag image (P41)](https://www.wikidata.org/wiki/Property:P41), which will likely make the query faster.

Screenshot_2020-04-06 Scholia

Daniel-Mietchen:
The solution to the question as to why https://tools.wmflabs.org/scholia/topic/Q3450985 (neurodevelopmental disorders) did not yield any result was simply that there were too few topic annotations (15 right now).

Screenshot_2020-04-06 Wikidata Query Service

I am starting a [first batch](https://tools.wmflabs.org/quickstatements/#/batch/28518) towards changing that.

egonw:
I checked the `tox` report and there are no problems.

fnielsen:
I am unfamiliar with the `target:` notation.

Daniel-Mietchen:
@fnielsen See #431 for the ```target:``` notation.

PR #1124, opened 2020-05-11
Added Bioschemas annotation for the Taxon aspect

egonw:


egonw:
@fnielsen, what do you think of the design? It doesn't need additional API calls, tho is does have to do some additional checks. I follow the design of elsewhere of wrapping in try/catch and a placeholder in the specific aspect `.html`.

PR #1134, opened 2020-05-22
Added a short tagline to the front page

egonw:
The tagline can be changed and is a suggestion.

Suggested by @chris-evelo

### DIFF
--- a/scholia/app/templates/index.html
+++ b/scholia/app/templates/index.html
@@ -7,7 +7,12 @@
 
 
 <div class="container">
-  Scholia is a service that creates visual scholarly profiles for topic, people, organizations, species, chemicals, etc
+  Scholia is a service that creates visual scholarly profiles for
+  <a href="topic/">topic</a>,
+  <a href="author/">people</a>,
+  <a href="organization/">organizations</a>,
+  <a href="taxon">species</a>,
+  <a href="chemical">chemicals</a>, etc
   using bibliographic and other information in Wikidata.
 </div>
 
--- a/scholia/app/templates/index.html
+++ b/scholia/app/templates/index.html
@@ -7,6 +7,11 @@
 
 
 <div class="container">
+  Scholia is a service that creates visual scholarly profiles for topic, people, organizations, species, chemicals, etc
+  using bibliographic and other information in Wikidata.
+</div>
+
+<div class="container">
   <div class="alert alert-warning" role="alert">
     Scholia relies on Wikidata, and Wikidata contains only a limited albeit growing subset of the corpus of scholarly literature, its authors and citations.
     Read more about the limitations in the <a href="faq">FAQ</a>.


Daniel-Mietchen:
Good idea. Perhaps we should link the listed aspects?

egonw:
I had considered that but was unsure: by linking out, the may do that (not itself bad), but they would miss out on the rest of the front page (maybe also not bad). So, added them now. Let me know what you think.

Chris-Evelo:
I still miss a general description both on the first page and even on the "about" page that sent me to background reading explaining what it is. I then found this in the paper abstract: 

"Scholia is a tool to handle scientific bibliographic information
through Wikidata."

"The Scholia Web service creates on-the-fly scholarly
profiles for researchers, organizations, journals, publishers, individual
scholarly works, and for research topics. To collect the data, it queries the
SPARQL-based Wikidata Query Service. Among several display formats
available in Scholia are lists of publications for individual researchers and
organizations, publications per year, employment timelines, as well as coauthor and topic networks and citation graphs."

I think that would serve the purpose.

BTW Egon, I think that is a law as old as the web that pages that link out more and in a useful manner actually get more, not less, traffic.

Chris-Evelo:
I also looked at the examples on the page. they make clear what I can do with Scolia. But as when looking at that for the first time I have no clue how to actually perform these queries. I think there should be links to the queries themselves, ideally in an editable environment where I can also search another name, author etc, and then execute the query.

egonw:
> BTW Egon, I think that is a law as old as the web that pages that link out more and in a useful manner actually get more, not less, traffic.

Yes, but these are not links out.

egonw:
> I also looked at the examples on the page. they make clear what I can do with Scolia. But as when looking at that for the first time I have no clue how to actually perform these queries. I think there should be links to the queries themselves, ideally in an editable environment where I can also search another name, author etc, and then execute the query.

People are not supposed here to edit the queries themselves. Do they can. Most panels have a "Edit on Wikidata" (or something like that) in the bottom right corner. More than open for discussion to me.

Chris-Evelo:
> > I also looked at the examples on the page. they make clear what I can do with Scolia. But as when looking at that for the first time I have no clue how to actually perform these queries. I think there should be links to the queries themselves, ideally in an editable environment where I can also search another name, author etc, and then execute the query.
> 
> People are not supposed here to edit the queries themselves. Do they can. Most panels have a "Edit on Wikidata" (or something like that) in the bottom right corner. More than open for discussion to me.

OK, I missed that. That is exactly what I meant. But it is a bit hidden, and I am probably not the only one to miss that. Maybe you can put a link behind every example that just says "Query" and goes there?

Daniel-Mietchen:
So here is how this looks like in Egon's draft:
Screen Shot 2020-06-08 at 22 03 44

I fiddled with that a bit further:
Screen Shot 2020-06-08 at 22 31 37

Once we agree on how things should look like on the front page, we have to adapt the About and FAQ pages accordingly.

Couldn't figure out how to issue a pull request on your branch, so I set up a new branch here in the repo: https://github.com/fnielsen/scholia/tree/patch/infoOnFrontPage-1134 .

fnielsen:
I have not add some other styling

Issue #1141, opened 2020-05-29
On topic profiles, add neighbourhood graph of the concept


Daniel-Mietchen:
[...]

PR #1142, opened 2020-05-29
added a neighbourhood graph for topic aspect, as per #1141

Daniel-Mietchen:


egonw:
I like the idea. See my comment on #1141, which then gives this for drugs:

image

Daniel-Mietchen:
What I am not sure about is the direction of the arrows, which is sometimes counterintuitive. As far as I am aware, it is determined automatically, and I am not aware of a way to define it in the query.

egonw:
IIRC, the direction of the arrows is based on the what is subject and what is object. Playing with this can be tricky, but you can swap direction by negating properties links in the SPARQL.

Daniel-Mietchen:
Turns out it's the [order of the output variables](https://www.wikidata.org/w/index.php?title=Wikidata:SPARQL_query_service/Wikidata_Query_Help/Result_Views&type=revision&diff=1203316150&oldid=1166098598&diffmode=source) that determines the arrow direction.

PR #1150, opened 2020-06-08
InChIKey not found page

egonw:
Implements https://github.com/fnielsen/scholia/issues/787

### DIFF
--- a/scholia/app/templates/404_chemical.html
+++ b/scholia/app/templates/404_chemical.html
@@ -1,6 +1,7 @@
 {% extends "base.html" %}
 
-{% block page_content %}
+{% block scripts %}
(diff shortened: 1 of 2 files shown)


fnielsen:
It is indeed pretty slow. Perhaps the design could also be improve. I will merge it anyway.

egonw:
Yeah, there are now 1M InChIKeys in Wikidata. It's suffering from the same problem as articles now. My local scripts are broken too now. I need to find a solution too. Please keep me posted if you think of something (will do the same thing!).

egonw:
I found a faster query, see https://github.com/fnielsen/scholia/pull/1171

PR #1164, opened 2020-06-18
Added a table with identifiers

egonw:
Implementation for https://github.com/fnielsen/scholia/issues/297


fnielsen:
This is fine, but the GRIN URL has problems https://www.wikidata.org/wiki/Q12024  http://127.0.0.1:8100/taxon/Q12024

egonw:
Ah, interesting... this is because the link is not a formater URL thing... but the full URL is the statement value.

PR #1171, opened 2020-07-05
Faster query

egonw:
Brings down the query time from around 40 seconds to just under 15 seconds.


fnielsen:
Tests: 
* http://127.0.0.1:8100/inchikey/QTBSBXVTEAMEQO-UHFFFAOYSA-N
* http://127.0.0.1:8100/inchikey/QTBSBXVTEAMEQO-UHF

PR #1248, opened 2020-11-05
Use the oAuth disambiguator, per issue #1242

### LINKED ISSUE #1242: Update all links to the Author Disambiguator
We are currently linking to the Author Disambiguator via links to pages like
https://author-disambiguator.toolforge.org/?doit=Look+for+author&name=Yang%20Yang
that go to publicly visible pages that are rate-limited.

The Author Disambiguator also has pages like
https://author-disambiguator.toolforge.org/names_oauth.php?doit=Look+for+author&name=Yang%20Yang
that are not publicly visible, require OAuth authentication and are not rate limited.

We should probably link to the latter, since these outgoing links are meant for curation, which would involve editing and thus authentication.

egonw:
fixes #1242

### DIFF
--- a/scholia/app/templates/author_missing.html
+++ b/scholia/app/templates/author_missing.html
@@ -9,7 +9,7 @@
 SELECT 
   (COUNT(?work) AS ?count) 
   ?author_name 
-  (CONCAT("https://author-disambiguator.toolforge.org/?doit=Look+for+author&name=", 
+  (CONCAT("https://author-disambiguator.toolforge.org/names_oauth.php?doit=Look+for+author&name=", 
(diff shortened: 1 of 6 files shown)


Daniel-Mietchen (commented):
This misses to make the change in the [FAQ file](https://github.com/fnielsen/scholia/blob/ed21cc7cacb62ae9532ff3545006e88b05b7fb3d/scholia/app/templates/faq.html#L32).

egonw:
I am not sure it should be changed there.

Daniel-Mietchen:
> I am not sure it should be changed there.

You're right &mdash; no change needed there. Sorry. Looks ready for merge.

PR #1254, opened 2020-11-11
Update for Wikidata now including the InChI= prefix #BioHackEU20

egonw:


fnielsen:
@egonw I presume that the change of InChl has no effect on "Related compounds"?

egonw:
It should not. That works with the InChiKey.`

PR #1311, opened 2020-12-20
Fixed matching of classes

egonw:
Fixes https://github.com/fnielsen/scholia/issues/1310


fnielsen:
Now running, e.g., https://scholia.toolforge.org/event/Q50355797

PR #1313, opened 2020-12-21
404 page specific for unfound DOIs

egonw:
Similar to the input request in https://github.com/fnielsen/scholia/pull/1312 (IMPORTANT: the first patch is the same) and also using `citation-js`, this page introduces a 404 page for the `/doi/` resolver functionality (for when the DOI is not in Wikidata), and then uses `citation-js` and the DOI to generate QuickStatements. Feedback welcome!

image

### DIFF
--- a/scholia/app/templates/404_doi.html
+++ b/scholia/app/templates/404_doi.html
@@ -1,0 +1,13 @@
+{% extends "base.html" %}
+
+{% block page_content %}
+
+There was no work found with DOI <a href="https://doi.org/{{ doi }}">{{ doi }}</a>.
+
+<script src="{{ url_for('static', filename='js/citation.js') }}"></script>
+
+<h2>QuickStatements</h2>
+
+<pre class="citation-js" data-input="{{ doi }}" data-output-format="quickstatements"></pre>
+
+{% endblock %}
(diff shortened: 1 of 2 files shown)


egonw:
@egonw: Check why `https://doi.org/10.17169/refubium-5313` is looked up correctly but doesn't return quickstatements.

egonw:
Fixed upstream. When a new plugin release is made, I will create a new bundle ad update this PR. (Tho this PR does not have to wait for that)

egonw:
@Daniel-Mietchen suggested to have a button to link directly to QuickStatements. I think this is a good idea, and crafting an URL is possible. 

However...

the current implementation uses the citation.js Replacer and the HTML now looks like:

```html
<pre class="citation-js" data-input="{{ doi }}" data-output-format="quickstatements"></pre>
```

The page needs to be rewritten first to use `Cite` without the replacer so that we have access to the QuickStatements output. That can then be URL encoded and added to the URL pattern, see https://www.wikidata.org/wiki/Help:QuickStatements#Running_QuickStatements_through_URL

I suggest we do this rewrite at a later stage.

Daniel-Mietchen:
I just opened #1351 for that QuickStatements button, so we can exclude that part from this PR.

egonw:
@fnielsen, can you please let me know what the next step is here?

egonw:
Patch is merged in.

PR #1315, opened 2021-01-09
External SPARQL for /organization/ and /organizations/

egonw:


scholia/app/templates/organization_missing.html
```diff
@@ -1,148 +1,13 @@
...
-	 }
-     );
-     sparqlToDataTable(
-	 missingAuthorItemsSparql, "#missing-author-items",
-     );
- });
-
-
-</script>
+{% set aspect = "organization_missing" %}
```
fnielsen, line 4: I think this should be "organization-missing"

PR #1318, opened 2021-01-17
No biological pathways; they skew the statistics too much

egonw:
This has been frustrating me for too long now. I know it's a bit ad hoc, but then again, we do not have that many bots using P50 for data entries, so I can it can be justified.


fnielsen:
Do we have examples on where it is a problem? "they skew the statistics too much"

fnielsen:
This has been supersed by #1335

PR #1323, opened 2021-01-24
Add bioRxiv route

cthoyt:
This PR adds a route `/biorxiv/<biorxiv_id>` to the application similar to its function for arXiv. Since they work almost identically besides the Wikidata property used, some of the code was abstracted into helper functions and reused for both.

### DIFF
--- a/scholia/app/views.py
+++ b/scholia/app/views.py
@@ -13,7 +13,7 @@ from ..rss import (wb_get_author_latest_works, wb_get_venue_latest_works,
 from ..arxiv import metadata_to_quickstatements, string_to_arxiv
 from ..arxiv import get_metadata as get_arxiv_metadata
 from ..query import (arxiv_to_qs, cas_to_qs, atomic_symbol_to_qs, doi_to_qs,
-                     github_to_qs,
+                     github_to_qs, biorxiv_to_qs,
                      inchikey_to_qs, issn_to_qs, orcid_to_qs, viaf_to_qs,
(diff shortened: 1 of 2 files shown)


egonw (approved):
Passes the `tox` testing and functionality works for me. Approved. Intuitively, I'd just use the DOI, tho, but that's no reason not to have this.

fnielsen:
Do we have a test example?

cthoyt:
`2020.07.15.204701` should work

fnielsen:
I am trying to understand:
* 127.0.0.1:8100/biorxiv/028522 redirects correctly.
* http://127.0.0.1:8100/biorxiv/2020.07.15.204701 not found
* http://127.0.0.1:8100/biorxiv/204701 not found

fnielsen:
> I am trying to understand:

Now I see the example is not there "Graphein - a Python Library for Geometric Deep Learning and Network Analysis on Protein Structures" is not in Wikidata.

cthoyt:
@fnielsen thanks for considering this PR :)

fnielsen:
If the biorxiv is not in Wikidata, then I suppose one can construct a DOI that can be set to Magnus Manske. Alternatively, we could get the data directly from bioRxiv.

cthoyt:
I'm not sure how the service is deployed, but if there multiple workers, it might be okay to have it hang while negotiating with the bioRxiv API, sending to Wikidata, sleeping a bit to avoid race conditions, then querying Wikidata

PR #1324, opened 2021-01-26
Update /missing/ to /curation/

egonw:


egonw:
Okay, that flake error is unexpected. I did run the checks locally :(

Daniel-Mietchen:
@Egonw Instead of having multiple pieces of code doing the redirect for one of the aspects that have a /curation code, wouldn't it be better to have some ASPECTVARIABLE for this kind of URL pattern, such that it works for all aspects?

```
@main.route('/ASPECTVARIABLE/' + q_pattern + '/missing')
def show_author_missing(q):
    """Redirects to the new HTML rendering for missing information.
    Parameters
    ----------
    q : str
        Wikidata item identifier.
    Returns
    -------
    html : str
        Redirected HTML.
    """
    return redirect(url_for('app.show_' + ASPECTVARIABLE + '_curation', q=q), code=301)

```

A similar optimization might probably be useful for the /curation pages themselves too.

egonw:
@fnielsen, is that possible? I can see how I should add an `aspect_pattern` but it not entirely clear to me how that would create an `aspect` parameter in the `show()` function.

Also, I think this is outside the scope of this patch. If we want to go this way, then we have a lot more code to rewrite.

Daniel-Mietchen:
>> @egonw Instead of having multiple pieces of code doing the redirect for one of the aspects that have a /curation code, wouldn't it be better to have some ASPECTVARIABLE for this kind of URL pattern, such that it works for all aspects?

> @fnielsen, is that possible? I can see how I should add an `aspect_pattern` but it not entirely clear to me how that would create an `aspect` parameter in the `show()` function.
> 
> Also, I think this is outside the scope of this patch. If we want to go this way, then we have a lot more code to rewrite.

I agree that this URL pattern rewrite is outside the scope of this patch, so I opened a new ticket for that &mdash; see #1326 .

Daniel-Mietchen:
@egonw When I tried your redirection code a few days ago, I could not get it to run, so I just gave it another try.
The curation pages work fine, but when I replace "curation" in a URL with "missing" to test the redirect, I am getting "connection refused" errors.

To reproduce, try https://gitpod.io/#https://github.com/fnielsen/scholia/pull/1324 , then go to a curation page and replace the "curation" in its URL with "missing".

egonw:
Looks like a Gitpod problem: it forgets to replace to replace the `localhost` in the URL with the URL of the pod/workspace.

Daniel-Mietchen:
I agree this looks like a Gitpod problem, so should not impede the merging of this patch.

fnielsen:
Why is there a conflicting file on scholia/app/templates/organization_missing.html

PR #1325, opened 2021-01-30
Added ROR identifier for organizations

egonw:
image

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -86,7 +86,6 @@ sparqlToIframe(`# tool: scholia
 	 }
 	 catch(e) { }
 
-	 console.log(item.claims.P6782[0])
 	 try {
 	     var rorID = item.claims.P6782[0].mainsnak.datavalue.value;
 	     detailsList.push( 'ROR <a href="https://ror.org/' + encodeURI( rorID ) + '">' + rorID + '</a> ');
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -90,9 +90,7 @@ sparqlToIframe(`# tool: scholia
 	     var rorID = item.claims.P6782[0].mainsnak.datavalue.value;
 	     detailsList.push( 'ROR <a href="https://ror.org/' + encodeURI( rorID ) + '">' + rorID + '</a> ');
 	 }
-	 catch(e) {
-	     console.log("ERROR: " + e)
-	 }
+	 catch(e) { }
 
 	 try {
 	     var hashtag = item.claims.P2572[0].mainsnak.datavalue.value;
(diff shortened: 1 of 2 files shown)


fnielsen:
Regarding "https://pbs.twimg.com": I do not think we should include thirdparty assets

egonw:
Agreed. I already asked them to upload their logo to WikiCommons.

cthoyt:
@egonw why not add other identifiers (e.g., GRID) while you're making this change? I guess one reason I could imagine why not is that Scholia is opinionated towards displaying only the most high quality annotations

fnielsen:
1. We shouldn't include external (non-wikimedia) assets.
2. It would be of interest to streamline things regarding inclusions of external identifiers.
3. having a ROR identifier there is fine, if we do not include external asset.

fnielsen:
Sorry for not being clear here: I will not accept PRs that include non-wikipedia assets (I also think that is against Toolforge rules, though not sure).

scholia/app/templates/base.html
```diff
@@ -86,6 +86,18 @@
 	 }
 	 catch(e) { }
 
+	 console.log(item.claims.P6782[0])
```
fnielsen, line 89: Why are there this console log message. Isn't it only for debugging?

scholia/app/templates/base.html
```diff
@@ -86,6 +86,18 @@
...
+	 try {
+	     var rorID = item.claims.P6782[0].mainsnak.datavalue.value;
+	     detailsList.push( '<img alt="ROR icon" src="' +
+			       'https://pbs.twimg.com/profile_images/1086592528038707200/qVDdvqOH_400x400.jpg' +
+			       '" width="16" height="16" hspace="4" /> <a href="https://ror.org/' +
+			       encodeURI( rorID ) + '">' + rorID + '</a> ');
+	 }
+	 catch(e) {
+	     console.log("ERROR: " + e)
```
fnielsen, line 98: Would this report an error if the external ID is not there?

egonw: Yes, I guess it would. You did not want that, right? I'll remove it.

scholia/app/templates/base.html
egonw: Yes, I guess it would. You did not want that, right? I'll remove it.

egonw:
And I'll remove the logo. It's hard readable anyway.

scholia/app/templates/base.html
fnielsen: Yes, we should not have console.logs here. Unless there is something wrong.

egonw: sorry, wanted to remove that one already. Done now.

scholia/app/templates/base.html
egonw: Removed

PR #1329, opened 2021-02-01
Update setup.py

### LINKED ISSUE #1328: Update Python version in setup.py from 2.7 to 3.5
as per
https://github.com/fnielsen/scholia/blob/172ed891f59342e6cdeec899eb4a59c43a46aa69/setup.py#L23
and
https://github.com/fnielsen/scholia/issues/952 .

Daniel-Mietchen:
fixes #1328


fnielsen:
I suppose we could have both? The tox test is actually for both?

Daniel-Mietchen:
@fnielsen I think we should at least add 3.5 (and probably think about upgrading from that too to something more recent) but dropping 2.7 would seem to make sense too, since we have not find any remaining 2.7 dependencies.

egonw:
I've been running Scholia with 3.x for a long time now.

fnielsen:
Toolforge is running 3.7.3, so that version should be tested

fnielsen:
As far as I remember there can be multiple lines and they should somehow correspond to what is in tox.ini

egonw:
@fnielsen, I'm developing with Python 3.7.3 and can confirm it works well.

egonw:
I am not a `tox.ini` wizard, but I have something like:

```
[tox]
envlist = flake8, pydocstyle, py37


[testenv:py27]
commands =
    python -m pytest --doctest-modules scholia
    python -m pytest tests
deps=
    pytest
    -rrequirements.txt

[testenv:py37]
commands =
    python3 -m pytest --doctest-modules scholia
    python3 -m pytest tests
deps=
    pytest
    -rrequirements.txt

[testenv:flake8]
commands = flake8 scholia
deps =
    flake8
    flake8-docstrings

[testenv:pydocstyle]
commands = pydocstyle --convention=numpy scholia
deps =
    pydocstyle
```

PR #1334, opened 2021-02-05
Fix #1331: academic age compute on academic works

### LINKED ISSUE #1331: "academic age" should only include academic works
The current query also takes into account non-academic works. For example, it makes my academic age 32, while this should be 20/21 instead.

fnielsen:
Previous SPARQL query only query all works to compute academic age.
Now only scholarly articles are queries. Perhaps other forms of works
should also be added to the query, e.g., academic book, though these
might be difficult to identify.

### DIFF
--- a/scholia/app/templates/work.html
+++ b/scholia/app/templates/work.html
@@ -185,8 +185,10 @@ WHERE {
   OPTIONAL {
     SELECT ?author_ (MAX(?academic_age_) AS ?academic_age) {
       wd:{{ q }} wdt:P50 ?author_ ;
-                   wdt:P577 ?publication_date .
-      ?author_ ^wdt:P50 / wdt:P577 ?other_publication_date .
+                 wdt:P577 ?publication_date .
+      [] wdt:P31 wd:Q13442814 ;
+         wdt:P50 ?author_ ;
+         wdt:P577 ?other_publication_date .
       BIND(YEAR(?publication_date) - YEAR(?other_publication_date) AS ?academic_age_)
     }
     GROUP BY ?author_


fnielsen:
Changed the SPARQL query a bit, see, e.g., https://scholia.toolforge.org/work/Q63433973 It will probably not be estimated for all kinds of researchers.

RichardNevell:
This is very interesting.

It occurs to me that academic journal article (Q18918145) is used as an instance of for nearly 500 items. I've never understood the distinction between that and scholarly article. Perhaps the two should be merged, or Q18918145 could be added to the query?

There are also nearly 500 instances of academic chapter (Q21481766) which could perhaps be added to the query.

What do you think about adding theses since they are academic work?

How to pick up on academic books looks like a tougher challenge.

fnielsen:
I think a scholarly article is a generic catch all that entails academic/scholarly(?) journals as well as academic/scholarly conference, worshop and symposium articles. Preprint could - I imagine - also be used. Academic change could very well be added. I suppse the reason for only scholarly articles was to avoid more expensive queries. I think they well could be added. Academic books are surely more difficult. I suppose it may be possible to catch them via the the publisher.

Do you have good examples where the academic age is poorly estimated?

RichardNevell:
It makes sense to keep the query lightweight, which prompted my thought on whether scholarly article and academic journal article should be merged (or at least the different usages made a bit clearer).

Because most of the information Scholia uses is about scholarly articles, the are very examples where the academic age is poorly estimated. I've spotted a couple of cases where someone has written a book or thesis before an article, but haven't been keeping track much. There is this one which returns a negative value: https://scholia.toolforge.org/work/Q59226576

Because it's a book review ([Q637866](https://www.wikidata.org/wiki/Q637866)) as a P31 rather than P136 it's not getting picked up as a scholarly article.

fnielsen:
If the academic age is estimated from the first scientific article then I suppose it make sense to have a negative academic age. Mikkel Wallentin is an example of an academic author that wrote fiction before his scientific work, e.g., https://scholia.toolforge.org/work/Q21076992. So I wouldn't be so concern about negative ages.

I have looked at https://www.wikidata.org/wiki/Q5054574 that is set to academic journal article. I suppose we can change that if we want consistency. I do not recall whether this has been discussed before.

PR #1361, opened 2021-03-06
Improvements to the CiTO experience

egonw:
Include:

- now lists the more reused articles in a journal (based on 3 CiTO intention types)
- introduces a CiTO button for journals and works
- some textual guidance


egonw:
Example for the CiTO button:

image

Note that it introduces a new feature: buttons for subpages. The current buttons only switch to another aspect. This can also be used for the curation pages. A SPARQL query is used so that the button only shows up if there is information to show, similar to the "normal" aspect buttons.

egonw:
This is what the new "growth over time" on the main /cito page looks like:

image

PR #1362, opened 2021-03-07
Report the taxons a compound is found in (with source)

egonw:
New in the `chemical` aspect:

image


fnielsen:
What is the example?

fnielsen:
> What is the example?

I found an example. There are apparently benzen in species!

fnielsen:
Maybe there should be an indication of the DOI going outside Scholia

egonw:
> Maybe there should be an indication of the DOI going outside Scholia

Did you create a ticket for that? We have a lot of links out across the aspects.

fnielsen:
> Did you create a ticket for that? We have a lot of links out across the aspects.

No

PR #1365, opened 2021-03-08
Add initial ChemRxiv support

cthoyt:
Similarly to #1323, this PR adds support for using the newly minted ChemRxiv property (P9262) as a lookup.

Example:  ChemRxiv identifier [13607438](https://chemrxiv.org/articles/preprint/Enantioselective_Conjugate_Addition_of_Catalytically_Generated_Zinc_Homoenolate/13607438) points to [Q104931192](https://www.wikidata.org/wiki/Q104931192). There aren't a ton of identifiers pre-populated, but I have written a script with the Wikidata Integrator that can add the rest.

### DIFF
--- a/scholia/app/templates/work_empty.html
+++ b/scholia/app/templates/work_empty.html
@@ -68,7 +68,7 @@ Scientific articles, conference articles, books, ...
               <dt>
                   <a href="../arxiv/1507.04180">arxiv/1507.04180</a>
               </dt>
-              <dd>Redirect based on a arXiv identifier.</dd>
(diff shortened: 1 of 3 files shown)


egonw (changes requested):
Looks good to me and `tox` finds no issues. One suggestion, can you add a rewrite example in `work_empty.html` in the third column? 

image

cthoyt:
@egonw solved in df14b33. I could imagine a more general/automated solution that reduces boilerplate code, but I'm worried that will get a bit too complicated.

Screen Shot 2021-03-08 at 11 37 14

egonw (approved):

PR #1406, opened 2021-03-29
Added a <meta> element for Unpaywall support

egonw:
It works by adding a `<meta>` element, which is the "semantic" data needed by the Unpaywall extension. See https://support.unpaywall.org/support/solutions/articles/44002172370-how-do-i-make-the-unpaywall-extension-work-on-my-site-

In action:

image

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -101,7 +101,7 @@ sparqlToIframe(`# tool: scholia
 
 	 try {
 	     var doi = item.claims.P356[0].mainsnak.datavalue.value;
-	     detailsList.push( '<meta name="citation_doi" content="' + doi + '"/>');
+	     $("head").append( '<meta name="citation_doi" content="' + doi + '"/>' );
 	 }
 	 catch(e) {}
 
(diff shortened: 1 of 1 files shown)


scholia/app/templates/base.html
```diff
@@ -99,6 +99,12 @@
...
 	 catch(e) {}
 
+	 try {
+	     var doi = item.claims.P356[0].mainsnak.datavalue.value;
+	     detailsList.push( '<meta name="citation_doi" content="' + doi + '"/>');
+	 }
+	 catch(e) {}
+
 	 try {
 	     $( '#details' ).append( detailsList.join( " | " ) );
```
fnielsen, line 109: Why is this not added to the head instead, like https://stackoverflow.com/questions/11018638/using-jquery-to-add-metadata-in-the-head-tag ?

fnielsen: `$("head").append(<meta name="citation_doi" content="' + doi + '"/>');` ?

egonw: Wasn't needed for the plugin to pick it up. But sounds good to me. I did not know the `$("head")` functionality.

egonw: done.

PR #1453, opened 2021-06-06
Added retraction info

egonw:
It uses the same three different ways Wikidata currently can mark an article as retracted (two in one UNION clause). The output looks like:

image


egonw:
Okay, the query works in the WDQS, but not within Scholia...

egonw:
In Scholia, I get this error:

image

egonw:
On using POST with WDQS: https://phabricator.wikimedia.org/T112151

PR #1456, opened 2021-06-08
Added 'dataset' as aspect

egonw:
Ammar, Daniel, can I get some initial feedback from you, please? What additional info should be listed for datasets?

Screenshot:

image


fnielsen:
Is the interesting person for a dataset necessarily an author. Don't we have curation, editor and developer as possible for datasets too?

egonw:
Agreed! I am not sure how far we are with [CRediT](http://credit.niso.org/) in Wikidata :)

PR #1461, opened 2021-06-19
Adds the 'Submit to QuickStatements' button

egonw:
Implements https://github.com/fnielsen/scholia/issues/1458

### DIFF
--- a/scholia/app/templates/404_doi.html
+++ b/scholia/app/templates/404_doi.html
@@ -1,13 +1,40 @@
 {% extends "base.html" %}
 
+{% block scripts %}
+{{super()}}
+
+<script src="{{ url_for('static', filename='js/citation.js') }}"></script>
+<script>
+const c = require('citation-js')
+try {
+  let example = new c.Cite([ '{{ doi }}' ])
+  let output = example.format('quickstatements')
+  $( '#qs' ).append( output );
+  output = encodeURI(
+    output.replaceAll('\t', '|')
+          .replaceAll('\n', '||'))
+          .replaceAll('/','%2F')
+  htmlOutput = "<a href=\"https://quickstatements.toolforge.org/#/v1=" + output + "\">\n" +
+               "  <button class=\"btn btn-primary\">Submit to Quickstatements ↗</button>\n" +
+               "</a>\n"
+  $( '#qs' ).after( htmlOutput );
+} catch (error) {
+  if (error.message.includes("status code 404")) {
+    $( '#qs' ).append( "DOI does not exist" )
+  } else {
+    console.log(error)
+  }
+}
+</script>
+
+{% endblock %}
+
 {% block page_content %}
 
 There was no work found with DOI <a href="https://doi.org/{{ doi }}">{{ doi }}</a>.
 
-<script src="{{ url_for('static', filename='js/citation.js') }}"></script>
-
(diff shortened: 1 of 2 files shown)


Daniel-Mietchen:
The code looks good, and I just created https://www.wikidata.org/wiki/Q107308161 while testing it.

scholia/app/templates/404_doi.html
```diff
@@ -1,13 +1,40 @@
...
+try {
+  let example = new c.Cite([ '{{ doi }}' ])
+  let output = example.format('quickstatements')
+  $( '#qs' ).append( output );
+  output = encodeURI(
+    output.replaceAll('\t', '|')
+          .replaceAll('\n', '||'))
+          .replaceAll('/','%2F')
+  htmlOutput = "<a href=\"https://quickstatements.toolforge.org/#/v1=" + output + "\">\n" +
+               "  <button class=\"btn btn-primary\">Submit to Quickstatements ↗</button>\n" +
```
Daniel-Mietchen, line 18: ```Quickstatements``` should be ```QuickStatements```.

scholia/app/templates/404_doi.html
```diff
@@ -1,13 +1,40 @@
...
+    console.log(error)
+  }
+}
+</script>
+
+{% endblock %}
+
 {% block page_content %}
 
 There was no work found with DOI <a href="https://doi.org/{{ doi }}">{{ doi }}</a>.
```
Daniel-Mietchen, line 34: I suggest to change this to 
```We found no work in Wikidata with the DOI <a href="https://doi.org/{{ doi }}">{{ doi }}</a>. If you would like to create the item for this work, you can use the <a href="https://www.wikidata.org/wiki/Help:QuickStatements">QuickStatements</a> tool by clicking the "Submit to QuickStatements" button (this requires authentication via <a href="https://www.mediawiki.org/wiki/Help:OAuth">OAuth 2.0</a>).```

Daniel-Mietchen (changes requested):
Works fine, but presentation could be improved.

scholia/app/templates/404_doi.html
egonw: Good catch. Sorry I overlooked that.

scholia/app/templates/404_doi.html
egonw: Sounds good to me.

fnielsen:
Is there a need of `urlencode`?

egonw:
> Is there a need of `urlencode`?

I copied that from the arxiv example and assumed so.

fnielsen:
arXiv has \d+.\d+ pattern I think. I don't know if that is checked. Can there be '"' in a DOI? '<' and '>' might not be a problem.

egonw:
I've seen many interesting characters in DOIs. The regex in identifiers.org is `^(doi\:)?\d{2}\.\d{4}.*$`

fnielsen:
So I suppose we need an urlencoding?

egonw:
Yes, I think https://github.com/WDscholia/scholia/pull/1461/files#diff-89c1152879b643f0f5eb07072531c16873bb0581597c328b6efa86fddbd72c79R13 should stay.

fnielsen:
It is line 34 when it is not found.

PR #1464, opened 2021-06-21
Update 404_doi.html

Daniel-Mietchen:
as per https://github.com/fnielsen/scholia/pull/1461#pullrequestreview-688584806


egonw (approved):

scholia/app/templates/404_doi.html
```diff
@@ -31,7 +31,7 @@
 
 {% block page_content %}
 
-There was no work found with DOI <a href="https://doi.org/{{ doi }}">{{ doi }}</a>.
+We found no work in Wikidata with the DOI <a href="https://doi.org/{{ doi }}">{{ doi }}</a>. If you would like to create the item for this work, you can use the <a href="https://www.wikidata.org/wiki/Help:QuickStatements">QuickStatements</a> tool by clicking the "Submit to QuickStatements" button (this requires authentication via <a href="https://www.mediawiki.org/wiki/Help:OAuth">OAuth 2.0</a>). Thank you for helping to curate this information.
```
fnielsen, line 34: `"https://doi.org/{{ doi }}">` -> `"https://doi.org/{{ doi | urlencode }}">` ?

egonw: No, there is should not be needed.

fnielsen: What I am/was concern about was something like: https://scholia.toolforge.org/doi/%22%3Ea%3C/a%3E%3Cscript%3Ealert('hallo');%3C/script%3E%3Ca%3E This seems not to "work".

egonw: Ah, you mean input sanitation... that's not what I was thinking of when you were wondering about `urlencode` which is normally for other purposes. Yes, of course, input checking is a good point.

fnielsen: It is not input sanitation, but output sanitation :) The DOI can - I presume - contain a wide array of characters, so we need to convert the special character before they are outputted to the HTML

egonw: really not sure about that... if the DOI is valid, then the created URL is too. The only thing is to make sure the input DOI is valid.

fnielsen: The issue is if the DOI contains a quote character. Then if might very well be a valid DOI but it makes the HTML invalid.

egonw: Okay, your suggested edit two days ago is saying something different, but I now think you want this: https://github.com/WDscholia/scholia/commit/a6f867646a59228291ab9f212772d2be93940c19 (added to this branch)

egonw: Because to me `"https://doi.org/{{ doi | urlencode }}"` is really about sanatizing the input.

fnielsen:
`<a href="https://doi.org/{{ doi | urlencode }}">{{ doi }}</a>` I think it is. The `{{ doi }}` is autoescaped (I very much hope). I hope that `{{ doi | urlencode }}`will escape the '"'.

PR #1471, opened 2021-06-28
Simplify running Scholia locally

cthoyt:
This PR does the following to make it easier to develop Scholia locally:

1. Adds documentation in the README on how to install Scholia locally
2. Adds an entrypoint in the setup.py file so you can run `scholia` from the shell instead of `python -m scholia`
3. Adds a `run` command to the CLI that mimics `runserver.py`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,6 +5,21 @@
 
 Scholia is a python package and webapp for interaction with scholarly information in Wikidata_.
 
+Installation
+------------
+Scholia can be installed directly from GitHub with:
+
+.. code-block:: shell
+
+    $ python3 -m pip install git+https://github.com/WDscholia/scholia
+
+It can be installed in development mode with:
+
+.. code-block:: shell
+
+    $ git clone https://github.com/WDscholia/scholia
+    $ cd scholia
+    $ pip install --editable .
 
 Webapp
 ------
--- a/README.rst
+++ b/README.rst
@@ -34,6 +34,11 @@ The webapp displays scholarly profiles for individual researchers, research topi
     
 The information displayed on the page is only what is available in Wikidata.
 
+Run locally after installing with ``pip``:
+
+.. code-block:: sh
+
+    $ scholia run
 
 Script
 ------
--- a/scholia/__main__.py
+++ b/scholia/__main__.py
@@ -4,6 +4,7 @@ Usage:
   scholia arxiv-to-quickstatements [options] <arxiv>
   scholia orcid-to-q <orcid>
   scholia string-to-type <string>
+  scholia run
 
 Options:
   -o --output=file  Output filename, default output to stdout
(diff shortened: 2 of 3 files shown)


fnielsen:
I am a newbee to the entry point trick. Would it install a script if it is installed in a virtualenv? I cannot see that?

cthoyt:
Yes it would install a script if you're in a venv that's localized to the venv itself. The entrypoint itself creates a small exectutable python file that 

For example on another project `pykeen` where I have installed it in a virtualenv:

Screen Shot 2021-06-28 at 10 52 13

If you try running `pykeen` when the virtualenv isn't activated, you will get some funny results w.r.t. imports

fnielsen:
Hmmm... I have no `scholia` in my `~/venvs/scholia/bin/`

cthoyt:
You will have to `pip install` one more time (while the `scholia` venv is activated), since this configuration is interpreted by `pip` and `setuptools`. Let me know if that doesn't address your issue

cthoyt:
wherever your virtualenv is, should be on the path so your console recognizes the commands inside it. Are you using a virtualenv / what's the path to it?

cthoyt:
Worst case scenario, at least `python -m scholia run` works, right?

fnielsen:
Hmmm... There is still something wrong
```
$ which python
/home/<user>/venvs/scholia/bin/python
(scholia) ~/venvs$ python -m scholia run
Usage:
  scholia arxiv-to-quickstatements [options] <arxiv>
  scholia orcid-to-q <orcid>
  scholia string-to-type <string>
```

cthoyt:
@fnielsen I will add a flag to get the current git version to see if we're running exactly the same code

fnielsen:
Issue #1633 with package_data should now be fixed with #1651

fnielsen:
This have been merged https://github.com/WDscholia/scholia/commit/764ffd876db456877d2536ed41d6c64c75f9d5d9

cthoyt:
@fnielsen I don't understand the way you went about merging the branch. It removed the history and attribution, and also did not update the pull request to denote that the code was merged. Would it be possible to merge it through here instead?

cthoyt:
@fnielsen thanks!

fnielsen:
@cthoyt I neither understood the way I went. The problem was that I wanted to have the `find_packages` into the PR, but then also found out that I also need `package_data` set. This was done in another PR. I got confused about the branching then. Now I have reopen the PR and pressed rebase and merge. I am wondering if the attribution is now there? I am looking at https://github.com/WDscholia/scholia/commits/master and https://github.com/WDscholia/scholia/network The commit you did now  shows up in the master line. I wonder if the were missing before?

fnielsen:
BTW I can install Scholia now and do `scholia run`. I think `unset PYTHONPATH` was an issue.

cthoyt:
Ah okay, a rebase or merge from the main branch would have done that. I think GitHub has a setting for enforcing PRs are rebased against master before merging to make sure this isn't an issue.

fnielsen:
I did a `git rebase master` locally before merging and pushing:

```bash
git checkout cthoyt-simplify-launch 
git pull git://github.com/cthoyt/scholia.git simplify-launch
less setup.py 
git rebase master
less setup.py
git checkout master
git merge --no-ff cthoyt-simplify-launch
less setup.py 
git push origin master
```
It is unclear to me why the PR was dangling on the remote GitHub.

cthoyt:
the issue is you pushed back to master rather than pushing back to cthoyt/simplify-launch, then merge the commit via github

cthoyt:
I think there's a setting in GitHub to make sure that you have push privileges even to external repos that make a PR back to this one. Right now I see this box isn't checked by default (even though I think for any good OSS project it should be)

PR #1477, opened 2021-07-04
Added 'stop' unicode to retraction info in work data

egonw:


fnielsen:
@egonw I have already merged it but wouldn't it be better with ⛔

egonw:
ah, there is more than one stop sign characters?

Issue #1189, opened 2020-09-07
Okay, in a discussion with @ammar257ammar we found the reason why Google's test tools do not recognize the Bioschemas annotation on this website: it is blocked by Wikidata :(


AlasdairGray:
Okay, in a discussion with @ammar257ammar we found the reason why Google's test tools do not recognize the Bioschemas annotation on this website: it is blocked by Wikidata :(

![image](https://user-images.githubusercontent.com/26721/92116402-12b6e280-edf4-11ea-87fb-42a15eb58c44.png)

cc @AlasdairGray

_Originally posted by @egonw in https://github.com/fnielsen/scholia/issues/923#issuecomment-686464429_

AlasdairGray:
Sorry, was trying to generate this in a different repo

PR #1478, opened 2021-07-04
Add Bioschemas by proxying Wikidata content (making Google bots happy)

egonw:
@fnielsen, this is now a finished patch, but if you have additional ideas, plz let me know (if not, please do merge in).

The new design solve the problem of the `robots.txt` blocking calls and limiting the SEO indexing of Scholia pages:

- a Scholia proxy is defined with the URL pattern `/$qid/bioschemas` which returns JSON
- the existing `base.html` uses this new call instead of a call to wikidata.org (with the `robots.txt` problem)

The extra call is only made when the aspect template has an `id=bioschemas` holder.

Possibly future optimization:

- [x] `property_for_q()` calls are replaced by a single `properties_for_q(q, {"P235": "key1", "Pxx": "key2", ...})`
- [x] 2-3 helper functions get added to simplify the code
- [x] other bits of the page get included in a similar way (like descriptions), making it also available for SEO
- [x] use the Wikidata description as Bioschemas content

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -87,6 +87,11 @@ sparqlToShortInchiKey(`# tool: scholia
 
 <script type="text/javascript">
     {% if q %}
+    if ( $( "#bioschemas" ).length ) {
+      var bioschemasurl = "{{ url_for('app.index') }}{{ q }}/bioschemas";
+      $.get(bioschemasurl, function(data) { $( '#bioschemas' ).append( data) } )
+    }
+
     var url = 'https://www.wikidata.org/w/api.php?action=wbgetentities&ids=' +
 	       '{{ q }}' + 
 	       '&format=json&callback=?';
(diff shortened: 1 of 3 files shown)


egonw:
@fnielsen, yes, I see the flake issues. I will fix those, when you told me you like the idea.

scholia/app/views.py
```diff
@@ -130,6 +131,117 @@ def redirect_q(q):
...
+
+    # get the aspect specific bits
+    if entity_class == "author":
+        data = {
+            "@type": "Person",
+            "http://purl.org/dc/terms/conformsTo": {
+                "@type": "CreativeWork",
+                "@id": "https://bioschemas.org/profiles/Person/0.2-DRAFT-2019_07_19/",
+            },
+            "description": "A person",
```
cthoyt, line 153: @egonw consider using the wikidata description for the person?

scholia/app/views.py
```diff
@@ -130,6 +131,117 @@ def redirect_q(q):
...
+            "@type": "MolecularEntity",
+            "http://purl.org/dc/terms/conformsTo": {
+                "@type": "CreativeWork",
+                "@id": "https://bioschemas.org/profiles/MolecularEntity/0.5-RELEASE/",
+            },
+        }
+        inchi_key = property_for_q(q, "P235")
+        if inchi_key:
+            data['inChIKey'] = inchi_key
+        if inchi:
```
cthoyt, line 191: it seems I may have lost the query for the inchi string

scholia/app/views.py
egonw: yes, I have :) The description goes for most of the items, but I really first need to know if @fnielsen likes the idea of proxying before I start tweaking stuff

scholia/app/views.py
egonw: no worries. I want to add `properties_for_q(q, ["P235", "Pxx"])` first anyway (return a map)

cthoyt: i was just thinking about how that would be nice!

fnielsen:
Should `/$qid/bioschemas` return HTML? And with content negotiation you can get json?

egonw:
> Should `/$qid/bioschemas` return HTML? 

No, the intention is only JSON.

> And with content negotiation you can get json?

Yeah, I guess that would be more correct: only return JSON when asked for JSON, return 415 for any other media type.

So, go ahead?

egonw:
Okay, I will assume this is a go ahead.

scholia/app/views.py
egonw: the `inchi` is now fetched, but still in the old way.

scholia/query.py
```diff
@@ -809,6 +809,42 @@ def q_to_label(q, language='en'):
...
+    url = 'https://query.wikidata.org/sparql'
+    params = {'query': query, 'format': 'json'}
+    response = requests.get(url, params=params, headers=HEADERS)
+    data = response.json()
+
+    results = data['results']['bindings']
+    if len(results) == 1:
+        return results[0]['value']['value']
+    else:
+        return None
```
cthoyt, line 845: If there are 2 or more results, it will return none. This will fix that:

```suggestion
    if not results:
        return None
    else:
        return results[0]['value']['value']
```

scholia/app/templates/base.html
```diff
@@ -87,6 +87,11 @@
 
 <script type="text/javascript">
     {% if q %}
+    if ( $( "#bioschemas" ).length ) {
+      var bioschemasurl = "{{ url_for('app.index') }}{{ q }}/bioschemas";
```
cthoyt, line 91: should be able to do the following, which knows how to automatically construct the URL based on the name of the function registered to the app blueprint (which is strangely named `main` and not `app`...)

```suggestion
      var bioschemasurl = "{{ url_for('app.bioschemas', q=q) }}";
```

It works because even though the endpoint name is constructed through semi-dynamic string concatenation, it still knows tha variable for the function is called `q`.

egonw: feel free to make PRs against this PR again :)

cthoyt: You should be able to directly accept/commit the suggestions from the github PR interface, which is effectively the same

cthoyt: I guess that lacks the benefit of me actually having to test it first 🙃

egonw: oh, lol, I did not know that! cool! I now see the button, but I guess I visually ignored it (pretty much like how my brain automatically ignores a lot of spam on webpages).

scholia/app/views.py
egonw: The method is now implemented: https://github.com/WDscholia/scholia/pull/1478/commits/3ef5ba659245a1dfbcc202f98f3cc9fbf6aac885

scholia/query.py
cthoyt: @egonw looks like this got clobbered

scholia/query.py
```diff
@@ -809,6 +809,103 @@ def q_to_label(q, language='en'):
...
+        if var in data['results']['bindings'][0].keys():
+            value = data['results']['bindings'][0][var]['value']
+            if prefixes is not None and prefixes[var]:
+                # add the given prefix for this Wikidata property
+                value = prefixes[var] + value
+            bskey = props[var]
+            # okay, some Bioschemas keys will have multiple values, in 
+            # which case the string type needs to be replaced by a list type
+            if bskey in bioschemas.keys():
+                if type(bioschemas[bskey]) == "list":
```
cthoyt, line 899: this won't do what you're expecting - `type()` returns a class, not a string. To get the desired effect, you could use:

```suggestion
                if type(bioschemas[bskey]) == list:
```

Or even better, using an instance check is more safe to class polymorphism (yes, this is a bit pedantic, but it's safer)

```suggestion
                if isinstance(bioschemas[bskey], list):
```

egonw: fixed. (Weirdly, the code was doing exactly what I wanted it to do...)

egonw:
Current tests:

- http://127.0.0.1:8100/Q12024/bioschemas
- http://127.0.0.1:8100/Q18216/bioschemas
- http://127.0.0.1:8100/Q20035886/bioschemas
- http://127.0.0.1:8100/Q21109365/bioschemas

These test normal behavior for 4 Bioschemas profiles, prefixing, and list/string variation.

egonw:
@fnielsen, it seems the `lamba` syntax is not available in Python 3.5 yet. Is that a blocker? Should I rewrite that?

scholia/query.py
egonw: The method is fully replaced now and I removed it.

scholia/app/views.py
```diff
@@ -135,6 +135,26 @@ def redirect_q(q):
     return redirect(url_for(method, q=q), code=302)
 
 
+def bioschemas_type(bioschemasType, conformsTo):
```
cthoyt, line 138: don't forget to use snake case!

scholia/query.py
egonw: @fnielsen, @cthoyt, I think I copied this from this line: https://github.com/WDscholia/scholia/blob/master/scholia/query.py#L806 which we may also want to fix (not part of this patch).

scholia/app/views.py
egonw: fixed.

scholia/app/views.py
```diff
@@ -134,6 +135,98 @@ def redirect_q(q):
     return redirect(url_for(method, q=q), code=302)
 
 
+def bioschemas_type(bioschemas_type, conforms_to):
```
cthoyt, line 138: The following suggestion adds type annotations (you'll need to add `from typing import Any, Mapping` at the top, too)

```suggestion
def bioschemas_type(bioschemas_type: str, conforms_to: str) -> Mapping[str, Any]:
```

scholia/app/templates/base.html
egonw: `app.bioschemas_for` fixed (patch pending)

scholia/app/views.py
egonw: descriptions are now added.

scholia/app/views.py
egonw: I'm going to mark this one for future updates. I don't think this is generally done yet, and can be done in a more general code base upgrade.

scholia/query.py
```diff
@@ -809,6 +809,107 @@ def q_to_label(q, language='en'):
...
+    Parameters
+    ----------
+    q : str
+        String with Wikidata Q item.
+    language : str
+        String with language identifier
+
+    Returns
+    -------
+    label : str
```
cthoyt, line 824: this documentation style is more appropriate for when you return a 2-tuple instead of a dictionary. Maybe using a `namedtuple` would even be best here!

scholia/query.py
```diff
@@ -809,6 +809,107 @@ def q_to_label(q, language='en'):
...
+
+    Examples
+    --------
+    >>> property_for_q('Q47512', {"P235": "inChIKey"})
+    {"inChIKey": "QTBSBXVTEAMEQO-UHFFFAOYSA-N"}
+
+    """
+    optionals = str('\n'.join(list(
+        map(lambda s: f"  OPTIONAL {{ ?item wdt:{s} ?{s} }}", props)
+    )))
```
cthoyt, line 878: This suggestion does 2 things:
1. It switches to using a list comprehension, which is _much_ more readable than a map/lambda
2. It fixes a logical bug. If you iterate over a dictionary `props`, it will iterate only over the keys. But it looks like you actually want to iterate over the items so you can use the key as the wdt:... and the value as the name in the query itself.
```suggestion
    optionals = '\n'.join(
        f"  OPTIONAL {{ ?item wdt:{key} ?{name} }}"
        for key, name in props.items()
    )
```

egonw: 1. PR welcome.
2. no, it should actually iterate only over the keys, not the values

scholia/query.py
egonw: ah, good catch. copy/paste artifact.

scholia/query.py
```diff
@@ -809,6 +809,107 @@ def q_to_label(q, language='en'):
...
+    variables = ' '.join(list(map(lambda s: f"?{s}", props.keys())))
```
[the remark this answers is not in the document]
egonw: done.

fnielsen:
Article on Bioschemas: https://scholia.toolforge.org/work/Q108345828

fnielsen:
This has been reverted. It was an accidental merge and there are at least two styling errors.

fnielsen:
Where is the issue for this PR?

egonw:
> Where is the issue for this PR?

Basically, this is the issue: https://github.com/WDscholia/scholia/issues/1189

egonw:
> This has been reverted. It was an accidental merge and there are at least two styling errors.

Can you give more detail, please? When I run tox, I get:

```
  pydocstyle: commands succeeded
```

PR #1482, opened 2021-07-05
Gitpod root URL - fixes #1327

### LINKED ISSUE #1327: Fix Gitpod root URL issue
as described in https://github.com/fnielsen/scholia/pull/1324#issuecomment-771036649 .

Daniel-Mietchen:
as per https://flask.palletsprojects.com/en/2.0.x/quickstart/#public-server

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -5,7 +5,9 @@ ports:
 - port: 8100
   onOpen: open-preview
 tasks:
-- init: >
+- before: |
+    export FLASK_APP=runserver.py
+  init: >
     python -m pip install -r requirements.txt 
   command: >
-    python runserver.py
+    flask run -h 0.0.0.0 -p 8100    
--- a/runserver.py
+++ b/runserver.py
@@ -7,4 +7,4 @@ app = create_app(
 app.config['APPLICATION_ROOT'] = '/'
 
 if __name__ == '__main__':
-    app.run(debug=True, port=8100)
+    app.run(debug=True, port=8100, host='0.0.0.0')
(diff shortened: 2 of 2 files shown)


fnielsen:
But isn't this a security issue? The Flask server should not by default be open to the public? If that is the case and we need it for something, I think we need an option.

fnielsen:
The title of the PR could be a bit more descriptive.

Daniel-Mietchen:
> But isn't this a security issue? The Flask server should not by default be open to the public? If that is the case and we need it for something, I think we need an option.



> But isn't this a security issue? The Flask server should not by default be open to the public? If that is the case and we need it for something, I think we need an option.

I agree this could potentially be a security issue, though not sure to what extent the Toolforge or Gitpod environments provide some shielding here.

Since the issue is mostly about using the Gitpod environment, maybe we should instead implement this via gitpod.yml, not runserver.py .

curibe:
@Daniel-Mietchen @fnielsen To implement this via gitpod.yml we can modify the file as follow:
```
image:
  file: .gitpod.Dockerfile

ports:
- port: 8100
  onOpen: open-preview
tasks:
- before: |
    export FLASK_APP=runserver.py
  init: >
    python -m pip install -r requirements.txt 
  command: >
    flask run -h 0.0.0.0 -p 8100
```
This only use 0.0.0.0 for gitpod enviroment, and It is not exposed to all public because the port still private:
image

fnielsen:
This seems more like it. 0.0.0.0 in the runserver.py would expose a developer running Scholia locally and this is removed now as I see it. Now only the gitpod server is using 0.0.0.0 and the port is also spedified.

fnielsen:
Do we have a Gitpod link i Scholia? https://gitpod.io/#https://github.com/WDscholia/scholia

fnielsen:
> Do we have a Gitpod link i Scholia?

Yes, it is in the README

PR #1485, opened 2021-07-07
plugin for search bar fix #1460

### LINKED ISSUE #1460: Styling and extending the search bar
Styling and extending the search bar in the menu and on the front page. Make it more similiar the search bar on Wikidata. 
* Add borders to the items.
* Add "more" option, so items beyond 10 can be selected.

DanielFLopez:
add select2 for search bar


fnielsen:
We cannot have thirdparty assets on the Wikimedia server, so cdn.jsdelivr.net is a nogo. At the moment we are serving static assets through flask and I suggest that for now. Wikimedia Foundation has some static servers, I we might look into that in the future.

scholia/app/templates/base.html
```diff
@@ -47,7 +47,8 @@
 
 <script type="text/javascript" src="{{ url_for('static', filename='jquery.dataTables.min.js') }}"></script>
 <script type="text/javascript" src="{{ url_for('static', filename='scholia.js') }}"></script>
-
+<link href="https://cdn.jsdelivr.net/npm/select2@4.1.0-rc.0/dist/css/select2.min.css" rel="stylesheet" />
```
fnielsen, line 50: The thirdparty asset is a problem on Wikimedia servers.

DanielFLopez: I add static files for select2

DanielFLopez:
Screen Shot 2021-07-08 at 9 05 10 AM

Issue #1479, opened 2021-07-04
Facilitate onboarding of new contributors


Daniel-Mietchen:
There are several components to this:
- good documentation both within the code as well as in the repo (e.g. README) and issues (e.g. labels)
- interoperability (e.g. indentation style)
- behavioural guidance (e.g. code of conduct or issue templates)

Some more points in [this post](https://dev.to/msaaddev/files-every-open-source-project-must-have-2mmm) (albeit with a certain bias for Node).

Daniel-Mietchen:
We already have 
- [README.rst](https://github.com/WDscholia/scholia/blob/master/README.rst)
- [LICENSE](https://github.com/WDscholia/scholia/blob/master/LICENSE)
- [label:good first issue](https://github.com/WDscholia/scholia/labels/good%20first%20issue)
- #941 for a Code of Conduct 
- #942 for issue templates
- #943 for pull request templates.

Daniel-Mietchen:
Here is some more guidance:
- [Ten simple rules for helping newcomers become contributors to open projects](https://doi.org/10.1371/journal.pcbi.1007296)
  - Rule 1: Be welcoming
  - Rule 2: Help potential contributors evaluate if the project is a good fit
  - Rule 3: Make governance explicit
  - Rule 4: Keep knowledge up to date and findable
  - Rule 5: Have and enforce a code of conduct
  - Rule 6: Develop forms of legitimate peripheral participation
  - Rule 7: Make it easy for newcomers to get started
  - Rule 8: Use opportunities for in-person interaction—With care
  - Rule 9: Acknowledge all contributions
  - Rule 10: Follow up on both success and failure

Issue #847, opened 2019-10-07
Look into regular releases


Daniel-Mietchen:
ideally with release notes and automated documentation

Issue #495, opened 2018-09-05
A production vs a beta version


dartar:
I find that many experimental features make it to the "official" Scholia instance incrementally and change often, somewhat breaking expectations about what information a user can find in each Scholia panel and in what order.

Shouldn't we have a separate public beta/staging environment where new features can be discussed and reviewed before being integrated into the main instance?

I personally feel many features are not "production ready" and Scholia is mature enough to distinguish a stable vs beta version, have formal feature announcements and releases etc.

fnielsen:
We could put up a beta environment. I do not mind being experimental on the official site though. Of course if users gets frustrated about features (missing or new unexpected) then it is a problem. I wonder if users know where to report problems? 

I am wondering whether we can point to some problems users have run into? Personally I feel a lot could benefit from moving away from embedded frames (relates e.g. to #301). Another issue is to add reasonator-like information. The response time is more and more going to be an issue. Which panels are relevant to show is also up for question. On each page we shouldn't have too many, - I think, due to the load on the WDQS. Some ways around it is by moving panels to subaspects or make the panels configurable.

PR #1486, opened 2021-07-07
Generate documentation and add link to readme (#1479)

carlinmack:
For #1479 

This PR introduces DOCUMENTATION.md which is generated by [pdoc](https://pdoc3.github.io/pdoc/). This lists all the docstrings in one place, allowing people to easily search for the modules and methods that may be of use to them. There are two caveats:

- Unfortunately the generator that was used (`pdoc`) only outputs in markdown, HTML and PDF. As such, DOCUMENTATION.md will be the only markdown file in this repository.
- `pdoc` does not correctly handle the `from .x import y` import format. As such, two modules — `qs` and `tex` — must be adjusted so that the documentation can be generated. Therefore these should be tested in the environments they are used, to ensure that these are not breaking changes. 

If either of these are deal breakers, I'll look into other generators.

With the addition of a config file, hyperlinking can be added to the module names ([example](https://github.com/carlinmack/afp/blob/main/DOCUMENTATION.md)) which is helpful for UX, but I didn't want to introduce [this file](https://github.com/carlinmack/afp/blob/main/src/pdoc/text.mako#L94) to the repository.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -685,6 +685,45 @@ Functions
 :   Write Pajek network file from SPARQL query.
 
 
+Module qs
+=========
+Quickstatements.
+
+Functions
+---------
+
+    
+`paper_to_quickstatements(paper)`
+:   Convert paper to Quickstatements.
+    
+    Convert a paper represented as a dict in to Magnus Manske's
+    Quickstatement format for entry into Wikidata.
+    
+    Parameters
+    ----------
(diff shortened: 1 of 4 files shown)


Daniel-Mietchen:
I had a first look and liked what I saw. We should probably couple this to official release workflows too, as per #495 and #847.

carlinmack:
closing in favour of #1557

Issue #1284, opened 2020-11-23
Check that all SPARQL queries have been converted to externalized query format


Daniel-Mietchen:
The goal here is to make sure that all SPARQL queries that were originally contained in .html pages have been moved to .sparql files and those .sparql files are called appropriately, as per https://github.com/fnielsen/scholia/pull/1267 .

Daniel-Mietchen:
Also check for queries that are commented out, as per https://github.com/fnielsen/scholia/issues/1170 .

Daniel-Mietchen:
Another thing to check here is that all queries have a proper HTML section ID, as per https://github.com/fnielsen/scholia/issues/1093 .

carlinmack:
With the above merged PRs, @curibe can this be closed?

fnielsen:
I supose. Do we have a "grep" that can check?

carlinmack:
Not sure if this is thorough but:

```sh
/scholia/scholia/app/templates$ grep SELECT *.html
base.html:    SELECT DISTINCT ?aspect
base.html:    SELECT DISTINCT ?aspectsubpage
```

carlinmack:
In the sidebar, click "Unpin this issue" 👍

curibe:
> Not sure if this is thorough but:
> 
> ```shell
> /scholia/scholia/app/templates$ grep SELECT *.html
> base.html:    SELECT DISTINCT ?aspect
> base.html:    SELECT DISTINCT ?aspectsubpage
> ```

@carlinmack These queries was handled in this [PR](https://github.com/WDscholia/scholia/pull/1538) because it has JS code, and It is only in the base.html file

Issue #1283, opened 2020-11-23
Convert topic aspects into external SPARQL query format


Daniel-Mietchen:
as per https://github.com/fnielsen/scholia/pull/1267

- [x] topic (Egon): https://github.com/fnielsen/scholia/pull/1424
- [x] topics
- [x] disease
- [x] taxon (Egon)
- [x] gene
- [x] protein
- [x] pathway (Egon)
- [x] lexeme
- [x] use (Egon)

This excludes chem* aspects, which are handled in https://github.com/fnielsen/scholia/issues/1270 .

egonw:
I'll take `taxon` because of https://github.com/fnielsen/scholia/issues/1300

Daniel-Mietchen:
All of these are functional now.

Issue #1282, opened 2020-11-23
Convert project aspects into external SPARQL query format


Daniel-Mietchen:
as per https://github.com/fnielsen/scholia/pull/1267

- [x] project
- [x] clinical-trial

Daniel-Mietchen:
Done.
![Screenshot 2021-08-02 at 05-30-17 Scholia](https://user-images.githubusercontent.com/465923/127839060-3a580e3f-7d5a-4267-b6a1-10b9db6fde3e.png)
![Screenshot 2021-08-02 at 05-31-10 Scholia](https://user-images.githubusercontent.com/465923/127839076-55905c3e-6da2-4309-bc1e-980522175f9a.png)

Issue #785, opened 2019-07-17
JavaScript development setup


rasmuserik:
Separating the JavaScript code from the html code, would make it easier to develop and test.

This would also make it easier to solve #390, by compiling code with backticks to a older JavaScript version supported IE.

Daniel-Mietchen:
Re separating JavaScript and HTML, see also this project, which did this to some extent already: https://code-openness.github.io/teamThree .

egonw:
@rasmuserik, I guess you mean the JavaScript that defines the SPARQL queries and runs them?

rasmuserik:
Self-assigned this, as I will look into the JavaScript setup during #1411.

PR #1487, opened 2021-07-07
check and externalize explicit sparql queries #1284

curibe:
related  #1283, #1282, #785 

## Externalize explicit sparql queries   

Converting the explicit queries (tables) first to externalized query format. For example, files who have the following queries:

```
somePanelDescriptionSparql = `
SELECT ...
...
`
``` 
This PR solves partially the issue #1284.

The following files still have explicit queries:

- [x] 404_chemical.html
- [x] author.html
- [x] author-index-curation.html
- [x] authors.html
- [x] award_curation.html
- [x] chemical-index-curation.html
- [x] chemical-index.html
- [x] lexeme_empty.html
- [x] pathway_empty.html
- [x] pathway.html
- [x] property.html
- [x] software_empty.html
- [x] topic_curation.html
- [x] use_empty.html
- [x] use.html
- [x] venue_curation.html
- [x] venues.html
- [x] work_cito.html
- [x] work_empty.html
- [x] works.html

### Some annotations
 - Files with empty will use the aspect as `index` but the name of html file will be the same for the moment. For example,
`use_empty.html` have an aspect named `use-index`. The name of these html will be change in other PR
- Externalization for `author.html`, `pathway.html` and `404_chemical.html` required the creation of  new macros and custom JS functions for extracting queries to external .sparql files


scholia/app/templates/author-index-curation.html
```diff
@@ -165,32 +33,32 @@ <h2>Author name strings to be resolved</h2>
...
 Wikidata items.
 Follow the link to use the Author disambiguator tool to try to resolve
 the authors.
 
-<table class="table table-hover" id="missing-citing-author-items"></table>
+<table class="table table-hover" id="missing-citing-author-items-table"></table>
 
 
-<h2>Authored works with missing topics</h2>
+<h2 d="missing-topic-header">Authored works with missing topics</h2>
```
DanielFLopez, line 59: maybe it's id?

fnielsen: yes, that should be id

curibe: Yes, It's id

fnielsen:
Sorry, about the name pattern of the files not being clear. It should be aspect_panel-description, e.g., `{% set aspect = "chemical-index" %}`. This is confusingly used. So the file is `chemical-index_statistics.sparql`

fnielsen:
`venues` are named fine!

fnielsen:
Files called "empty" should be "index" instead, e.g., "work-index" and "pathway-index". Sorry again

fnielsen:
`award_curation` -> `award-curation`

curibe:
> `award_curation` -> `award-curation`

@fnielsen In this case the html is called "award_curation.html". Do I still assign it in aspect as "award-curation"? no matter if the html file uses "_" or "-"?

curibe:
> Files called "empty" should be "index" instead, e.g., "work-index" and "pathway-index". Sorry again

@fnielsen Do you want to change the name only in the `{%set aspect=....%}` or also the html name?

fnielsen:
> > `award_curation` -> `award-curation`
> 
> @fnielsen In this case the html is called "award_curation.html". Do I still assign it in aspect as "award-curation"? no matter if the html file uses "_" or "-"?

All HTML and sparql files should be called with a dash instead of an underscore

curibe:
@fnielsen please, can you check this commit  11c714837b3934b10f2e884b9082503c843c9bae  where a new macro was implemented togetther with JS functions in `scholia.js`

curibe:
@fnielsen do you also require this caption here?
image

fnielsen:
> @fnielsen do you also require this caption here?

I am not sure what you mean?

curibe:
> > @fnielsen do you also require this caption here?
> 
> I am not sure what you mean?

I refer to the caption shown in the red box. The function that shows the diagram  does not a code implemented  to display the
link, but another caption (below-left in the image :  _This diagram is  showing..._ ):
https://scholia.toolforge.org/pathway/Q28031254
image

I put it to keep the format as the other externalizations. You can see the changes in this commit: a37c6004609e30597fab99868a2858dc726bcad1
The caption was added here:
image
And It looks like this:
image

fnielsen:
No a caption there is not necessary, I think. The rendering does not come from Wikimedia services, but from a thirdparty service, available at https://pathway-viewer.toolforge.org/?id=WP111 . I do not think they construct the visualization with SPARQL.

fnielsen:
* cito_empty should be cito-index
* is venues iframe panels missing? http://127.0.0.1:8100/venues/Q1981225,Q5936947

PR #1500, opened 2021-07-09
arxiv redirect to quickstatement fix #129

### LINKED ISSUE #129: If an arxiv does not exist redirect to quickstatement construction
If an arxiv does not exist redirect to quickstatement construction

DanielFLopez:
fix #129 
add a redirect for arxiv if does not exist

Screen Shot 2021-07-09 at 4 40 04 PM
Screen Shot 2021-07-09 at 4 39 45 PM

### DIFF
--- a/scholia/app/views.py
+++ b/scholia/app/views.py
@@ -1393,7 +1393,11 @@ def show_search():
 
     arxiv = string_to_arxiv(query)
     if arxiv:
-        return show_arxiv(arxiv)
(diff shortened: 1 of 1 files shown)


egonw (approved):
Is that common, that there is more than one result?

fnielsen:
> Is that common, that there is more than one result?

I do not think there should be more than one result. Perhaps in some rare cases with preprints.

PR #1503, opened 2021-07-12
Fix ellipses

carlinmack:
Currently a description will look like:

> Sentence one. Sentence two. Sentence three. ... (From Wikipedia)

This PR will instead have:

> Sentence one. Sentence two. Sentence three... (From Wikipedia)

Some more formal (legal, journal) sources say if the source ends in a period you must show this by having four periods, but I don't think we require such rigour for this description.


carlinmack:
Superceded by #1540

PR #1505, opened 2021-07-13
Add links to all panels 

### LINKED ISSUE #348: Provide way to link to individual panels in an aspect
ideally in a persistent way

carlinmack:
Close #348 

image

Users can hover over titles and simply click to get the link to that section in the URL bar. 

All H2s and H3s with ID's on all pages will have a link added to them

Followed guidance from https://amberwilson.co.uk/blog/are-your-anchor-links-accessible/ to ensure they are still accessible


fnielsen:
Should the titles be black instead of blue?

fnielsen:
Why isn't it black in my browser with: 
```css
.hlink,
.hlink:hover {
    color: black;
}
```

carlinmack:
hmm what browser are you using? Titles are black for me

carlinmack:
ah! If you control+f5 then the new CSS & JS will be loaded and it will display correctly. This can be avoided by [adding a version string to the source links](https://stackoverflow.com/questions/32414/how-can-i-force-clients-to-refresh-javascript-files), something that's probably possible with Flask

PR #1507, opened 2021-07-13
Use arXiv API

### LINKED ISSUE #43: Use arXiv API instead of HTML
Change `get_metadata` in arxiv.py to use API.

### LINKED ISSUE #1059: Arxiv scraping does not seem to work anymore
Arxiv scraping does not seem to work anymore. Perhaps export.arxiv.org should be used instead or the API. arxiv.org scraping works locally so perhaps server-side scraping from Toolforge has been turned off.

carlinmack:
Close #43 and close #1059

This is an exact replica of the previous version using the API instead

---

Message about the first commit:

> I looked into the API and they only serve [Atom](https://en.wikipedia.org/wiki/Atom_(Web_standard)) not JSON or anything familiar. They recommend using [feedparser](https://pypi.org/project/feedparser/). As we would be importing an external module any way, I looked into it and found the [arxiv](https://pypi.org/project/arxiv/) module. This makes things really easy and externalises the handling of the API to another project, great!

> Unfortunately as the name clashes with the file, and Python does not like importing things with the same name we either have to:
> [ ] Find a workaround for the Python import system which is robust (I looked into a lot of options, but I'm also not too familiar with this)
> [ ] Rename our module (which has a nice, fitting name)
> [ ] Implement the parsing using the feedparser module

> Will do the latter option tomorrow if I don't hear back.

> P.S. I've changed the default behaviour to return null for entries without doi's, which seems easier to handle to me. Reflecting I should probably just maintain the behaviour we currently have. Will also do this tomorrow.
image

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,4 +8,4 @@ requests
 simplejson
 werkzeug>=0.9
 urllib3>=1.25.1
-arxiv
+feedparser
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ requests
 simplejson
 werkzeug>=0.9
 urllib3>=1.25.1
+arxiv
--- a/scholia/arxiv.py
+++ b/scholia/arxiv.py
@@ -36,7 +36,7 @@ except ImportError:
 
 import requests
 
-from arxiv import Search
+from feedparser import parse as parse_api
 
 
 USER_AGENT = 'Scholia'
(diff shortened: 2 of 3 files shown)


fnielsen:
```
$ flake8 scholia
scholia/arxiv.py:18:1: F401 'dateutil.parser.parse as parse_datetime' imported but unused
scholia/arxiv.py:35:5: F401 'xml.etree' imported but unused
```

PR #1509, opened 2021-07-14
Add select2 to search results page

carlinmack:
From conversation in #1460

image

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -439,7 +439,11 @@ sparqlToIframe(`# tool: scholia
      let search_text = ""
      $('#searchterm').select2({
          allowClear: true,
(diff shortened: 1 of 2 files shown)


Daniel-Mietchen:
Looks good to me.

fnielsen:
Now the "/search" page has a self-link

carlinmack:
So there's kind of two searches possible from this page. The autocomplete is searching a different index to the "Advanced search". For example the results for "Tim" with the Advanced search (below) are different to the results in the auto-complete (above) 

image

I think this is where the confusion was in #1460. The self link in the search page is because the user might want to do the "Advanced search" and therefore go back to the results page. 

I agree this is currently confusing, will think of how they can be reconciled/made clearer to the user

PR #1510, opened 2021-07-14
Improve styling of search page

carlinmack:
Click to enlarge image 

image

### DIFF
--- a/scholia/app/static/css/scholia.css
+++ b/scholia/app/static/css/scholia.css
@@ -5,8 +5,10 @@
 }
 
 .search-results li {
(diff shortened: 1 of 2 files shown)


fnielsen:
* The "result" headline is small compared to the result lines
* The next back links are on top of each other. Should it be next to each other?

carlinmack:
* Realistically the heading is just a formality - think google (or anywhere else) search results, there isn't a header unless there are multiple different types of results.
* ~~I'm not sure what you mean?~~ oh you are correct haha, I will fix

PR #1527, opened 2021-07-17
Remove unnecessary page controls on Authors and other pages

carlinmack:
Missed panels in authors/ etc in the original PR (#1518)

image


egonw (approved):
Looks good to me and makes the UI less cluttered.

Issue #1508, opened 2021-07-14
DOI to quickstatement does not forward full quickstatements


fnielsen:
Example: 
- Go to https://scholia.toolforge.org/doi/10.1136/bmj.n1656
- Metadata is fetch ok
- Press "Submit to QuickStatements"
- Only the first QuickStatement line is shown (instance of).

I am not sure what happened. Have we changed anything in the code? Has quickstatement? Why do citation-js has a newline after CREATE? Is that a problem? The URL is https://quickstatements.toolforge.org/#/v1=%7CCREATE%7C%7C%7C%7C%7CLAST%7CP31%7CQ13442814%7CS248%7CQ5188229%7CS813%7C+2021-07-14T00:00:00Z%2F11%7C%7C%7CLAST%7CLen%7C%22The%20covid-19%20lab%20leak%20hypothesis:%20did%20the%20media%20fall%20victim%20to%20a%20misinformation%20campaign?%22%7C%7C%7CLAST%7CP304%7C%22n1656%22%7CS248%7CQ5188229%7CS813%7C+2021-07-14T00:00:00Z%2F11%7C%7C%7CLAST%7CP356%7C%2210.1136%2FBMJ.N1656%22%7CS248%7CQ5188229%7CS813%7C+2021-07-14T00:00:00Z%2F11%7C%7C%7CLAST%7CP407%7CQ1860%7CS248%7CQ5188229%7CS813%7C+2021-07-14T00:00:00Z%2F11%7C%7C%7CLAST%7CP577%7C+2021-07-08T00:00:00Z%2F11%7CS248%7CQ5188229%7CS813%7C+2021-07-14T00:00:00Z%2F11%7C%7C%7CLAST%7CP1433%7CQ546003%7CS248%7CQ5188229%7CS813%7C+2021-07-14T00:00:00Z%2F11%7C%7C%7CLAST%7CP1476%7Cen:%22The%20covid-19%20lab%20leak%20hypothesis:%20did%20the%20media%20fall%20victim%20to%20a%20misinformation%20campaign?%22%7CS248%7CQ5188229%7CS813%7C+2021-07-14T00:00:00Z%2F11%7C%7C%7CLAST%7CP2093%7C%22Paul%20D%20Thacker%22%7CP1545%7C%221%22%7CS248%7CQ5188229%7CS813%7C+2021-07-14T00:00:00Z%2F11%7C%7C%7C%7C

fnielsen:
[...]

PR #1530, opened 2021-07-19
Use a better encodeURI method

egonw:
Suggest in https://github.com/WDscholia/scholia/issues/1508


fnielsen:
Examples: https://orcid.org/0000-0001-6352-8580 https://doi.org/10.1007/978-3-030-43687-2_25

PR #1534, opened 2021-07-22
Add units to distance metric

carlinmack:
Depends on #1526 to be merged first. I think once it is merged, this PR will have only one commit

image

### DIFF
--- a/scholia/app/static/scholia.js
+++ b/scholia/app/static/scholia.js
@@ -241,7 +241,11 @@ function sparqlToDataTable2(sparql, element, filename, options={}) {
               if (i == 0) {
                 column['className'] = 'dt-right';
               }
-            } else if (column['title'] == 'Score' || /\Wper\W/.test(column['title'])) {
+            } else if (
+              column['title'] == 'Score' ||
+              column['title'] == 'Distance' ||
+              /\Wper\W/.test(column['title'])
+            ) {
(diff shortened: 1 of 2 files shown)


egonw (commented):
I like the idea. Some thoughts (but not sure what to do with it):

1. did you check for any other SPARQL result label named  Score/Distance? can we expect unexpected side effects?
2. can the number of digits be made customizable in the `sparqlToDataTable2()` call? (as variant function)
3. this distance is always in kilometre, right? do we want a full column for that?

fnielsen:
I am wondering whether it would be better to have the rounding in the SPARQL with, e.g., https://stackoverflow.com/questions/66296758

carlinmack:
These comments got lost, but to answer

@egonw:

1. This commit changed all score and distance columns to be rounded to 2dp
2. It could, it's not very often that this column appears though
3. The distance is always in km because of the `geof:distance(?other_geo, ?geo)` function (which from my research, the unit cannot be defined in the Wikidata version). The string "km" could be appended in the sparql, but then we couldn't round properly. It's probably possible to append "km" with DataTables, however we would need to be sure that all Distance columns are in kilometers


@fnielsen:
You'll notice that the original commit was being rounded with SPARQL
image
however you can see in the before image above, this does not always round to the same number of digits.

PR #1535, opened 2021-07-22
Highlight academic in academic tree

### LINKED ISSUE #438: In academic trees, highlight target person in separate color
Case in point: [Leibniz](https://tools.wmflabs.org/scholia/author/Q9047) (loads slowly)

carlinmack:
Close #438

image


egonw (approved):
Oh, nice one. Thanks!

(why did I not think of this?!?)

fnielsen:
If the person has no supervisor then it will not be colored.

PR #1537, opened 2021-07-22
Add HTML Titles to all pages

### LINKED ISSUE #524: HTML titles for SEO
I just noticed that all pages in Scholia use the same HTML title. Adding meaningful HTML titles using the item label is a simple fix that's guaranteed to drive traffic and discoverability. 

You could format them in the following ways:

## Label only

 - Scholia: Denny Vrandečić
 - Scholia: Technical University of Denmark

## Label and QID

 - Scholia: Denny Vrandečić (Q18618629)
 - Scholia: Technical University of Denmark (Q1269766)

## Facet and label

 - Scholia: Author: Denny Vrandečić 
 - Scholia: Organization: Technical University of Denmark

I'd personally favor 1. Label only or 3. Facet and label. The QID doesn't add anything meaningful to the end user. Exposing the facet in the HTML title could be useful for items that can have multiple aspects.

If you end up implementing it, check how it affects traffic data.

carlinmack:
Close #524

### DIFF
--- a/scholia/app/templates/404.html
+++ b/scholia/app/templates/404.html
@@ -1,5 +1,7 @@
 {% extends "base.html" %}
 
+{% block title %}Not Found - Scholia{% endblock %}
+
 {% block page_content %}
 
 Not found.
--- a/scholia/app/templates/author-index.html
+++ b/scholia/app/templates/author-index.html
@@ -1,5 +1,7 @@
 {% extends "base.html" %}
 
+{% block title %}Author - Scholia{% endblock %}
+
 {% block page_content %}
 
 <div class="container">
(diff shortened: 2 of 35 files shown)


carlinmack:
PS it's not immediately obvious to me what the difference between "empty" and "index" pages are. I feel like they should both be called "index".

curibe:
@carlinmack I think it is related to what Finn mentioned here: https://github.com/WDscholia/scholia/pull/1487#issuecomment-876777874

egonw (approved):
Looks good to me.

egonw:
@carlinmack, can you rebase on master instead of merge?

PR #1538, opened 2021-07-22
Separate js from html #785

curibe:
## Separate js codes from html

This PR basically separate a part of JS codes inside `base.html` file to get a more readable HTML file. The JS code extracted was written inside the file `scholia.js`

**Notes**
In `base.html`
 - Not all js code as extracted because it requires some jinja statements and variables
   
   image

 - The jquery statement $(document).ready is left inside html because It is used by the jinja macros
  
   image
 - The others js code are related with the externalization of sparql queries and it has been resolved in [PR 1487](https://github.com/WDscholia/scholia/pull/1487) and [PR 1517](https://github.com/WDscholia/scholia/pull/1517)


carlinmack:
Hi, I'm very in favour of this PR :) I would suggest however making workWithQ() into it's own file, and have:

```html
{% if q %}
<script src="{{ url_for('static', filename='workWithQ.js') }}"></script>
{% endif %}
```

This would help reduce page size (#1502) as not all pages need the full `scholia.js`. Thoughts?

curibe:
@carlinmack  something like this? :

```js
{% if q %}
        
       <script src="{{ url_for('static', filename='workWithQ.js') }}"></script>
	workWithQ("{{ q }}")

	{% if q2 %}
         // The same here
	  workWithQ2("{{ q2 }}")
	{% endif %}
	
	.
        .
        .
  {% endif %}
````

**Comment by carlinmack** [comment]
Hmm sorry just looking at `scholia.js` now, it seems that most of it is useful for working with Q-number pages? I thought it was more general. In that case it's probably better to go with this PR and then later on see if `scholia.js` can be dropped from the non-Q-number pages (like the home page).

File: scholia/app/static/scholia.js [hunk 1]
```diff
@@ -539,3 +539,419 @@ function sparqlToShortInchiKey(sparql, key,  element, filename) {
     sparqlToDataTable2(new_sparql, element, filename);
 
 }
+};
```
carlinmack, line 542: Did you mean to add this };?

scholia/app/static/scholia.js
```diff
@@ -539,3 +539,419 @@ function sparqlToShortInchiKey(sparql, key,  element, filename) {
...
+	 if ('en' in item.labels) {
+	     $("#h1").text(item.labels.en.value);
+	 }
+	 $("#h1").append(' (<a href="https://www.wikidata.org/wiki/' +  q  + ' ">' + q  + '</a>)');
+	 
+	 var detailsList = Array();
+	 try {
+	     var orcid = item.claims.P496[0].mainsnak.datavalue.value;
+	     detailsList.push( '<a href="https://orcid.org/"><img alt="' +
+			       'ORCID logo" src="{{ url_for(\'static\', filename=\'images/orcid_16x16.gif\') }}"' +
```
carlinmack, line 561: This might set this PR back awhile, but you can't use any `{{ }}` code in the extracted JS. This is because the html is rendered as a jinja template and these values are replaced before being executed by the browser.

I would suggest either dropping the template (as you can do here - replace it with "/static/images/orcid_16x16.gif") or in the case of dynamic stuff like `{{ q }}`, make it a parameter to the function.

carlinmack: PS, the above results in a `GET http://127.0.0.1:8100/author/%7B%7B%20url_for('static',%20filename='images/orcid_16x16.gif')%20%7D%7D 404 (NOT FOUND)` error in the console

scholia/app/static/scholia.js
```diff
@@ -539,3 +539,419 @@ function sparqlToShortInchiKey(sparql, key,  element, filename) {
...
+	 /* BioSchemas annotation */
+	 if (item.claims.P31 &&
+	     ((item.claims.P31[0].mainsnak.datavalue.value.id == 'Q5'))) {
+	   try { /* Person */
+	       bioschemasAnnotation = {
+	          "@context" : "https://schema.org",
+	          "@type" : "Person" ,
+	          "http://purl.org/dc/terms/conformsTo": { "@type": "CreativeWork", "@id": "https://bioschemas.org/profiles/Person/0.2-DRAFT-2019_07_19/" },
+	          "description" : "A person" ,
+	          "identifier" : "{{ q }}" ,
```
carlinmack, line 621: For example, this line can be changed to:
```js
"identifier" : q,
```
as `q` is passed as a parameter to this function

carlinmack (changes requested):
Will review the other changes once scholia.js is fixed :)

scholia/app/static/scholia.js
curibe: > This might set this PR back awhile, but you can't use any `{{ }}` code in the extracted JS. This is because only the HTML is rendered as a Jinja template, replacing these values before being executed in the browser.
> 
> I would suggest either dropping the template (as you can do here - replace it with "/static/images/orcid_16x16.gif") or in the case of dynamic stuff like `{{ q }}`, make it a parameter to the function.

Yeah, you right. I think you first recommendation is better because the GIF file is hard-coded

curibe: @carlinmack I used a hybrid between the two solutions in case you want to change the static path and also in the case you want to add more files from the static path

image

Please, let me know what do you think about that.

scholia/app/templates/base.html
```diff
@@ -84,457 +90,40 @@
...
-
-     // Add anchor links to all headings
-     var headers = document.querySelectorAll('h2[id], h3[id]')
+	 searchTerm(placeholder=`{{ 
+		 "Scientist, paper, topic, DOI, etc." if request.path.endswith("search") else "Search ..."
+		}}`,
+		urlfor="{{ url_for('app.index') }}")
+    
+	// Add anchor links to all headings
+	var headers = document.querySelectorAll('h2[id], h3[id]')
```
curibe, line 126: @carlinmack What do you want me to do with this code, leave it here or move it to the sholia.js file like the others?

scholia/app/templates/base.html
```diff
@@ -84,457 +90,40 @@
...
      if (headers) {
        headers.forEach(element => {
          var title = element.innerText;
```
curibe, line 129: @carlinmack What do you want me to do with this code, leave it here or move it to the sholia.js file like the others?

carlinmack: sorry for not getting back to you on this, I reckon we should try externalise all of the JS

carlinmack:
Could you have `function workWithQ(q, q2) {...}` so that q2 can be accessed inside `workWithQ`? For #1552, I need to have consecutive ajax calls 
```js
$.ajax(endpointUrl, settings).then(function (data) {
    .... stuff with 1 ...
}).then(function (data) { 
   if (q2) {
       $.ajax(endpointUrl, settings).then(function (data) {
            .... stuff with q2 ...
       })
   }
})
```
and therefore need to have access to q2

curibe:
> Could you have `function workWithQ(q, q2) {...}` so that q2 can be accessed inside `workWithQ`? For #1552, I need to have consecutive ajax calls
> 
> ```js
> $.ajax(endpointUrl, settings).then(function (data) {
>     .... stuff with q ...
> }).then(function (data) { 
>    if (q2) {
>        $.ajax(endpointUrl, settings).then(function (data) {
>             .... stuff with q2 ...
>        })
>    }
> })
> ```
> 
> and therefore need to have access to q2

@carlinmack sure, no problem. I can do it.

scholia/app/templates/base.html
curibe: No problem. Ok, working on it!

egonw:
@curibe, I tried rebasing your 5 patches on top of the latest master, but there are more merge conflicts than I can resolve :( I can try harder, but it will take me a few hours, because it diverged quite a bit :(

curibe:
Hi @egonw  and @fnielsen . Ok, I'm going to take a look at it. Probably there were several changes in the js code since the PR was created, which makes the conflicts appear. Especially because the PR extracts the code out of the html.

Issue #1245, opened 2020-11-05
Define a mechanism for Scholia queries to identify themselves to the Wikidata Query Service


Daniel-Mietchen:
e.g. so that WDQS can gather more granular statistics.

egonw:
But this is independent from the external queries, not?

Daniel-Mietchen:
It is independent of the externalization, but should of course work with the externalized queries.

egonw:
Is this just about setting the HTTP header `User-Agent`? See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

Daniel-Mietchen:
I think this would be the most parsimonious way of doing it.

PR #1540, opened 2021-07-23
Replace TextExtracts call with REST API

### LINKED ISSUE #1501: Paragraph breaks are removed, with no space substituted.
On https://scholia.toolforge.org/topic/Q936 I see text from the English Wikipedia, including "satellite navigation devices.Created by Steve Coast ".

On Scholia, there is no space between "devices" and "Created". 

On the Wikipedia article there is a paragraph break between "devices" and "Created".

carlinmack:
As discussed in #1501

This PR will close #1501 

image
image
image

API documentation https://en.wikipedia.org/api/rest_v1/

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -253,16 +253,13 @@ sparqlToIframe(`# tool: scholia
 	 /* English Wikipedia */
 	 if ('enwiki' in item.sitelinks) {
 	     var title = item.sitelinks.enwiki.title;
-	     var wikipediaApiUrl = 'https://en.wikipedia.org/w/api.php?' +
-				   'action=query&prop=extracts&exsentences=3&exlimit=1&exintro=1&' + 
-				   'explaintext=1&callback=?&format=json&titles=' +
+	     var wikipediaApiUrl = 'https://en.wikipedia.org/api/rest_v1/page/summary/' +
 				   encodeURIComponent(title);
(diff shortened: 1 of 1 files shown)


scholia/app/templates/base.html
```diff
@@ -253,19 +253,19 @@
...
-	     
-	     $.getJSON(wikipediaApiUrl, function(data) {
-		 var pages = data.query.pages;
-		 var text = pages[Object.keys(pages)[0]].extract + " ... "
-		 var html = "(from the <a href=\"" + wikipediaUrl + "\">English Wikipedia</a>)";
-		 $("#intro").text(text).append(html);
-	     }).fail(function(d, textStatus, error) {
-		 console.error("getJSON failed, status: " + textStatus + ", error: "+error)
+
+	     var headers = new Headers({"Api-User-Agent": "https://github.com/WDscholia/scholia"});
```
egonw, line 260: This is something we should use more often.

egonw (approved):
Looks good to me. Finn, what do you think?

scholia/app/templates/base.html
carlinmack: This is mentioned in #1245 , chose to use "Api-User-Agent" as there [is a issue with the "User-Agent" header in Chrome](https://stackoverflow.com/questions/42815087/sending-a-custom-user-agent-string-along-with-my-headers-fetch).

PR #1542, opened 2021-07-25
Add notice for preprints

### LINKED ISSUE #1180: Preprints should be clearly visible as preprint
We should make sure that preprints are clearly marked as such, with a notice that they have not been reviewed. Ideally, it would point to the Wikidata item of the published version (after which you could wonder if Wikidata should keep the item of the preprint, since we have so many publications anyway).

carlinmack:
Close #1180

Example Q56383735

image


egonw (changes requested):
The second text label is not correct. That should be something like "Published as" or "Peer-reviewed version" or "Preprint of"... I think I personally like the last best.

egonw (approved):
Thanks for the update!

fnielsen:
Example: 127.0.0.1:8100/work/Q56383735

PR #1549, opened 2021-07-28
Redirect to non-OAuth Author Disambiguator by default

carlinmack:
The user facing links should be changed to without-OAuth as author disambiguator has bad UX when you are not logged in

image

Compare with 

image

Will remove draft status when I thoroughly check the other links

### DIFF
--- a/scholia/app/templates/author_curation.html
+++ b/scholia/app/templates/author_curation.html
@@ -163,7 +163,7 @@ Missing information with respect to the author.
 <h2>Author name strings to be resolved</h2>
 
 The authorship may be represented as an author name string rather than a Wikidata item.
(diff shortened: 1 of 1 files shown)


Daniel-Mietchen:
There was a point when we had this the other way round, and I think we changed it because the primary users of this would usually be logged in on the Author Disambiguator.

Perhaps you can have a look at making the Author Disambiguator look nicer? The problem it has to deal with are rate limits, which can only be removed (at the moment at least) if users are logged in.

carlinmack:
If https://github.com/arthurpsmith/author-disambiguator/pull/160 gets merged, I'll make some more PRs there

PR #1551, opened 2021-07-28
Add CITATION.cff

fnielsen:
GitHub should now support citation support.


egonw:
@fnielsen, just a note, you will see that GitHub does not use the info yet:

image

For this, a version and data-released field need to be added.

fnielsen:
Should we cite v0.2?

egonw:
The most recent release sounds good to me.

Issue #741, opened 2019-06-13
Provide a mechanism to hide panels that do not have enough information


Daniel-Mietchen:
Along with useful default values.

Use case examples:
- page number panel on author aspect
- location panel in author aspect, as per https://github.com/fnielsen/scholia/issues/737

Potential solution: ASK queries, as per https://github.com/fnielsen/scholia/issues/617

PR #1552, opened 2021-07-28
Improve UX of aspect chooser

carlinmack:
50% of closing #1322 (the other half is the addition of a link to /curation)

    
The format of this mirrors the URL so hopefully users will figure out with use

image

With chooser open

image

On pages with a single aspect

image

On pages with a sub aspect

image

### Caveats

> Please list anything which has been left out of this PR or which should be considered before this PR is accepted
Check any of the following which apply:

I am not incorporating the "CiTO" subaspect as I aim to handle that in a different way

* [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ]  This change requires a documentation update
    * [ ]  I have made corresponding changes to the documentation
* [x]  This change requires new dependencies (please list)
 
From the Bootstrap documentation:

> Dropdowns are built on a third party library, Popper.js, which provides dynamic positioning and viewport detection. Be sure to include popper.min.js before Bootstrap’s JavaScript or use `bootstrap.bundle.min.js` / `bootstrap.bundle.js `which contains Popper.js. Popper.js isn’t used to position dropdowns in navbars though as dynamic positioning isn’t required.

We could use `bootstrap.bundle.min.js` instead of `popper.min.js` which is added by this PR?

If you make changes to the python code
  
* [ ]  my code passes [tox](https://tox.readthedocs.io/en/latest/) check, you can receive warnings about tests, documentation or both

### Testing
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

* Checked the links on pages with one, multiple and sub-aspects

### Checklist
* [ ] I have commented my code, particularly in hard-to-understand areas
* [x] My changes generate no new warnings
* [x] I have not used code from external sources without attribution
* [x] I have considered accessibility in my implementation 
    * I am assuming that the Bootstrap dropdown is accessible
* [x] There are no remaining debug statements (print, console.log, ...)

Issue #1322, opened 2021-01-18
Consistent design for navigating between pages


Daniel-Mietchen:
We currently have some consistency to navigate between aspect pages, but no consistent navigation between aspects and subaspects (see, e.g., #1244 ), nor to and from other subpages like /cito or /curation (cf. #1126).

This ticket is to collect notes on how to approach that. Today, we exchanged some more ideas about the curation pages specifically, e.g. considering a "Curate" button. 

Perhaps we could give it a different colour and position, to keep it apart from the "content" pages? On the other hand, the case could be made for keeping all the buttons together to highlight that contributing to curation is integral to the usage of Scholia.

The main problem I see with having the button on the top is that it might be irritating because when people are at the top of the page, they ain't seen nothin of the actual page content. So perhaps add it to the bottom, or to both paces, or have it slide as the user scrolls down?

Issue #374, opened 2018-05-14
Improve documentation about how to add/ curate information displayed by Scholia


Daniel-Mietchen:
A growing number of Scholia users are not familiar with how Wikidata works, so we have to help them find out.

PR #1556, opened 2021-07-29
Add links to curation pages

### LINKED ISSUE #281: Create some "missing" subpage for every aspect
E.g. right now we have 
https://tools.wmflabs.org/scholia/author/Q8219/missing
as a complement to 
https://tools.wmflabs.org/scholia/author/Q8219

but no 
https://tools.wmflabs.org/scholia/taxon/Q91703/missing
for
https://tools.wmflabs.org/scholia/taxon/Q91703 .

### LINKED ISSUE #1122: Give links to curation pages more prominence
E.g. by having them at the top and/ or bottom of the profile page, perhaps with some colour.

carlinmack:
2nd half of closing #1322 
Close #1122 

This PR adds a link to the curation page for all Q number pages. 

image

This PR relies on #1552 as the curation aspect chooser is filtered based upon whether there is a useful curation page for that aspect. On this curation page, the aspect chooser is filtered of the aspects which don't have curation page templates:

image

----

I'm happy with this PR now, text below preserved for reference

---

This needs some iteration/discussion before it's ready, due to design choices etc.

The position was chosen as users are accustomed to finding edit controls in the top right of the page. 

In terms of the text:

* "Edit" would have to direct the user to Wikidata, which would be odd
* "Curate" does make sense if you have the right mental model, but would be initially confusing to the user (especially with the current page, more on this below). 
* "Add more info" is clearer, but not ideal
* I've gone with "Improve data" as this doesn't have the connotation of directly editing

I have no comment on what the URL of the curation pages should be, I do prefer "curation" to "missing" 

> In terms of the curation page itself, I think it needs some work, definitely in regards to #374. As far as I can tell, this page largely functions to provide quick access to (hopefully) fruitful Author Disambiguator pages and Wikidata pages that have missing properties. As a user I would expect this page to provide some context of where information comes from, where it can be added, what tools are available, and _then_ links to pages on these tools.

> If we take the user story "As an academic, I want to add a notable paper to an academic's page, so that the Scholia profile is more complete", there is no direction for the best way to do something like this, and what tools are available to do this. 

> There is currently a hardcoded list of valid aspects for curation, but if we can close #281 then we do not need this.

> I will work on a prototype of this in another PR :)

PR has been created

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -87,9 +87,6 @@ sparqlToShortInchiKey(`# tool: scholia
 <script type="text/javascript" src="{{ url_for('static', filename='bootstrap-autocomplete.min.js')}}"></script>
 
 <script type="text/javascript">
-    var currentURL = window.location.pathname.split("/");
-    var currentAspect = currentURL[1];
-    
     {% if q %}
     var url = 'https://www.wikidata.org/w/api.php?action=wbgetentities&ids=' +
 	       '{{ q }}' + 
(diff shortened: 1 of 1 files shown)


carlinmack:
Oops, still need to handle this bit of code:

```js
// this query opens a subpage for the Wikidata item, e.g. as cito subpage
     var endpointUrl = 'https://query.wikidata.org/sparql';
     var query = `
    SELECT DISTINCT ?aspectsubpage
    WHERE {
    { [] pq:P3712 / wdt:P31 wd:Q96471816 ; ps:P2860 / wdt:P1433 wd:{{ q }} . BIND("venue-cito" AS ?aspectsubpage) }
    UNION
    { [] pq:P3712 / wdt:P31 wd:Q96471816 ; ps:P2860 wd:{{ q }} . BIND("work-cito" AS ?aspectsubpage) }
    }
    `;
     settings = {
	 headers: { Accept: 'application/sparql-results+json' },
	 data: { query: query }
     };

     $.ajax( endpointUrl, settings ).then( function ( data ) {
	 data.results.bindings.forEach(function(entry) {
	     var props = entry.aspectsubpage.value.split("-");
	     var aspect = props[0];
	     var subpage = props[1];
	     $( '#aspect-chooser' ).append(
		 `<a role="button"
		     class="btn btn-outline-dark btn-sm"
                     href="{{ url_for('app.index') }}` + aspect + '/' + '{{ q }}' + '/' + subpage + '">' + subpage + '</a> ' );
	 });
     } );
```

carlinmack:
As of #1610 , I will not handle that bit of code in this PR, and instead close #741 first

fnielsen:
I am trying to understand what should be done with this PR. Should I review and merge it? And the work can continue from there? Currently there is no automatic testing.

carlinmack:
Actually I think it is fine now, can't quite recall why this wasn't marked as ready for review

Daniel-Mietchen:
This also fixes #1708

Daniel-Mietchen (approved):
Looks good and works fine.

Daniel-Mietchen:
Adding a pointer to 
- #1807 

which serves to keep an overview of things we worked on during the hackathon.

PR #1562, opened 2021-07-30
Update DataTables and styling

carlinmack:
Fixes `Failed to load resource: the server responded with a status of 404 () - sort_both.png` error in the console on all pages with datatable panels

    
Initially I was wanting to reduce some of the warnings and errors in the web console while Scholia is running. Chiefly this appears on most pages of Scholia:

```
Failed to load resource: the server responded with a status of 404 () - sort_both.png
```

Investigating this, it turns out it's a Datatables file which is used to show the user they can sort. I start making a PR to add these images, but ~~then there's problems with the path to the images etc etc~~ there is a hardcoded URL in the CSS to `../images/sort_both.png`, and four other images for the different states. We would either have to do some flask trickery to change the path for these 5 images, or we would need to manually change the paths to `/static/images/X.png` each time we update the library. 

I then go to see if this has been fixed in recent versions, so I download the most recent version, but no. I spend some more time seeing if we should upgrade anyway (yes, there's a few security updates [one](http://cdn.datatables.net/1.10.19/) [two](http://cdn.datatables.net/1.10.22/)).

I then realise there's a [download builder](https://datatables.net/download/) that allows you to select a theme. There is a theme for our current version of Bootstrap (4), and it turns out it doesn't require sort_both.png. The tables also match the rest of our theme which is a big plus (I initially thought that the tables were also <iframe>s as they have such a different design language — corners, gradients and borders). As we can get the best of both worlds, I think we should use the Bootstrap 4 version of DataTables.

Before 

image

After

image

### Caveats

The only downside is that the tables are a little bit taller, but I think it's due to padding or something.

### Testing

* Checked a couple pages to ensure that the datatables still work as expected

### Checklist


fnielsen:
Isn't it possible just to put the images so they are served from "static"?

carlinmack:
Initially I did put the images there, however there is a hardcoded URL in the CSS to `../images/sort_both.png', and four other images for the different states. Every time we update the DataTables library (which we should due to the security issues), we would need to change the paths to `/static/images/X.png`. This could be achieved with a manual note, or a patch, etc. 

However, we can avoid this issue, and not serve images, _plus_ have the tables look more cohesive with the rest of the website*. As we can get the best of both, I think we should use the Bootstrap 4 version of DataTables.

* I initially thought that the tables were also <iframe>s as they have such a different design language (corners, gradients and borders)

PR #1570, opened 2021-08-04
Dockerize scholia

### LINKED ISSUE #1452: Scholia Docker image
Hello,

I was wondering if Scholia already has an official docker/container image. I have created one with the following configuration.

```Dockerfile
FROM python:latest

WORKDIR /app

COPY requirements.txt requirements.txt
RUN pip3 install -r requirements.txt

COPY . .

EXPOSE 8100

CMD [ "python3", "runserver.py"]
```

And modifying the `runserver.py` file as follows:
```python
app.run(debug=True, port=8100, host='0.0.0.0')
```
This image is available at docker hub as [wesogroup/scholia](https://hub.docker.com/repository/docker/wesogroup/scholia).

But, I don't know how I could tell Scholia in the runserver.py to start listening to all hosts without adding the `host=0.0.0.0` parameter in the `app.run(...)` method. Any idea?

I have been looking and I think the following issues might be related (https://github.com/fnielsen/scholia/issues/966, https://github.com/fnielsen/scholia/issues/1058). Are there any more updates on them other than what can be seen on GitHub?

Thank you very much.

curibe:
resolve #1452 

## Create a docker image for scholia

Scholia can be runned as a docker container using docker-compose. We also implement nginx as a container to work as a reverse proxy.

To run scholia with docker we proceed as follows:

```
# run scholia app only
docker-compose --profile local up

# run scholia with nginx
docker-compose --profile proxy up
```

### Some notes:

- To run flask in a container, you need to expose it to the host using 0.0.0.0 because if we use localhost, it would reference the container's localhost
- The host definition is done in the Dockerfile and not directly in the `runserver.py` file:  
image
- Despite using 0.0.0.0, this remains only for the host that is running docker and is not exposed if network accesses are configured.
- In case of limiting the IP, we can create a custom network in docker so that the flask app only receives requests from an IP in the subnet.


Daniel-Mietchen:
Looks good to me.

In terms of using the host ```0.0.0.0``` here, this seems quite similar to the approach taken for Gitpod in https://github.com/WDscholia/scholia/pull/1482 but I suppose these two things should be kept separate.

docker-compose.yml
```diff
@@ -0,0 +1,35 @@
+version: "3.9"
```
fnielsen, line 1: I am trying to understand the version here. I get "ERROR: Version in "./docker-compose.yml" is unsupported." while being on Docker version 20.10.7 and docker-compose version 1.25.0. 1.25.0 seems to be old based on information here https://docs.docker.com/compose/compose-file/ so I need to upgrade. I am on a 20.04 LTS system. Would it be possible to support older systems for the docker configuration files?

fnielsen:
These edits will get an old docker-compose working:
```
@@ -1,4 +1,4 @@
-version: "3.9"
+version: "3.3"
 
 services:
     # RUN SCHOLIA ALONE 
@@ -9,15 +9,11 @@ services:
             - 8100:8100
         volumes: 
             - .:/scholia
-        profiles:
-            - local
 
     # RUN SCHOLIA WITH NGINX
     webapp_proxy:
         image: scholia
         build: .
-        profiles:
-            - proxy
         volumes: 
             - .:/scholia
 
@@ -26,8 +22,6 @@ services:
         build: ./nginx/
         ports:
             - 80:80
-        profiles:
-            - proxy
         depends_on: 
             - webapp_proxy
```

PR #1580, opened 2021-08-09
Added panel to show in what biological process and protein complexes

egonw:
The new `/protein/Qxxx` aspect now looks like:

image


fnielsen:
@egonw I am wondering whether the "Complex" columns shouldn't be moved to /complex/ aspect?

fnielsen:
http://127.0.0.1:8100/protein/Q88656943

egonw:
> @egonw I am wondering whether the "Complex" columns shouldn't be moved to /complex/ aspect?

crap. yes, you're absolutely right. Need to think of the right column header...

PR #1584, opened 2021-08-09
Recognize protein complexes and proteins

egonw:


egonw:
Closely related to https://github.com/WDscholia/scholia/pull/1543

scholia/app/templates/base.html
```diff
@@ -395,17 +395,23 @@
...
+    UNION
     { wd:{{ q }} wdt:P50 | wdt:P2093 [] . BIND("work" AS ?aspect) }
     }
     `;
      settings = {
 	 headers: { Accept: 'application/sparql-results+json' },
 	 data: { query: query }
      };
-     
+     console.log(query)
```
fnielsen, line 409: Maybe you want to remove the console log?

scholia/app/templates/base.html
```diff
@@ -395,17 +395,23 @@
     UNION
     { wd:{{ q }} wdt:P644 [] . BIND("gene" AS ?aspect) }
     UNION
+    { wd:{{ q }} p:P31/ps:P31 wd:Q22325163 . BIND("complex" AS ?aspect) }
```
fnielsen, line 398: Why not wdt:P31?

fnielsen:
There were some extra commits here by me:

06c3b87aaed482fe7e1268786021b90340e0b2c2: Change path query to direct query

c9f50949ba507b9541dcd0ded600be3548ec0bf2: Made the complex type guesser broader

scholia/app/templates/base.html
egonw: Because when there is more than one statement with that property, then the RDF becomes less predictable.

fnielsen: I do not understand. Would multiple statements not resolve in multiple wdt's and muliple (p/ps)'s?

PR #1608, opened 2021-08-17
Add image for item inline with table of contents

### LINKED ISSUE #1599: Next to table of contents, add image for the item
The image size should probably scale in some useful way with the table of contents.

Here is a mockup for https://scholia.toolforge.org/topic/Q232169 .
Screen Shot 2021-08-12 at 19 18 30 copy

carlinmack:
Fix #1599

Add images to item pages inline with table of contents, making sure to resize appropriately. I used this comment to guide my approach, but didn't use any code https://stackoverflow.com/a/34402875

~~The `background-image` property is used as it's the only way to shrink an image to the size of it's neighbour without JS~~

image
image
image
    
### Caveats

~~For this to be accessible, the appropriate aria-roles should be placed so that it is clear that this is an image. Additionally, the `title=` attribute should be set so there is hover text and the image can be read out.~~ I briefly looked into getting the caption for the image from commons but it looked like I would need to make two API requests to get it. ~~When this `title` is added, there should be a fallback to the file name so that there is always a title set for the image.~~

* [x]  This change requires new dependencies (please list)
   * I needed to introduce a way to get the MD5 of a string as the URL includes the first 2 characters from the MD5 of the file name. I used the following implementation from stackoverflow https://stackoverflow.com/a/60467595 They note it as free to use. 
   
### Testing

* Checked on several pages as shown above.
* Also checked an example item without an image

### Checklist
* [x]  I have commented my code, particularly in hard-to-understand areas
* [x]  My changes generate no new warnings

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -134,7 +134,8 @@ sparqlToShortInchiKey(`# tool: scholia
       imageURL += imageName
       var itemImage = document.getElementById("item-image");
       if (itemImage) {
-        itemImage.style = "background-image:url('" + encodeURI(imageURL) + "')";
+        itemImage.src = encodeURI(imageURL);
+        itemImage.alt = imageName
       }
     }
 
@@ -616,12 +617,15 @@ sparqlToShortInchiKey(`# tool: scholia
 
     toc.appendChild(tocList)
 
-    var itemImage = document.createElement("div");
-    itemImage.id = "item-image"
+    var itemImage = document.createElement("img");
+    itemImage.id = "item-image";
+    itemImage.alt = "";
 
     tocParent.appendChild(toc);
     tocParent.appendChild(itemImage);
     document.querySelector("h2").insertAdjacentElement("beforebegin", tocParent)
+    
+    itemImage.height = toc.clientHeight
   }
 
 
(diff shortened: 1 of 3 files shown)


scholia/app/templates/base.html
```diff
@@ -124,6 +124,23 @@
     });
     {% endif %}
 
+    if ("P18" in item.claims) {
+      console.log("in")
```
DanielFLopez, line 128: I think you should remove the console.log

carlinmack: yep sorry, this was actually why I added the ` * [ ] I have removed debug statements` to the PR template :)

carlinmack:
Realised that as we are using JS anyway, we can set the height of the image correctly and we don't need to use the background image hack

egonw (changes requested):
Looks nice! Can you check http://127.0.0.1:8100/author/Q18618629 ? I guess something doesn't go well with the character encoding?

carlinmack:
It's actually an issue with the md5 function I chose. I've tried to comment on the answer to alert him of the problem but I need 50 upvotes on answers to be able to comment :/ I'll find a better implementation

carlinmack:
Fixed 👍 there are not really any good options for unicode supporting md5 functions, without adding a crypto library

PR #1616, opened 2021-08-19
Fix wrong link country topic 1547

### LINKED ISSUE #1547: Wrong links in "Authors of the country" panel for country-topic aspect
Example: https://scholia.toolforge.org/country/Q33/topic/Q2539

Both author and example work is wrong

DanielFLopez:
fix #1547

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,6 @@
 name: Python package
 
-on: [push]
+on: [push, workflow_dispatch]
 
 jobs:
   build:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,6 @@
 name: Python package
 
-on: [push, workflow_dispatch]
+on: [push]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


Daniel-Mietchen (approved):
Looks good and works fine.

fnielsen:
As far as I can see this add a "/" as the base of the directory. We previously had the root wmflabs.org/scholia/ which means that "/" was questionable. It is a question of whether this is a relic or we should maintain it. I think it should be possible to change "linkPrefixes" instead.

egonw:
> As far as I can see this add a "/" as the base of the directory. We previously had the root wmflabs.org/scholia/ which means that "/" was questionable. It is a question of whether this is a relic or we should maintain it. I think it should be possible to change "linkPrefixes" instead.

@fnielsen, can you clarify what the next step is. Is this a question? Does it need a change in the code? Who is going to take the next step? Stating such explicitly will help keep momentum, because people have clear what is expected from them.

fnielsen:
@egonw

> can you clarify what the next step is

I have now created an issue: https://github.com/WDscholia/scholia/issues/1700

fnielsen:
This is superceded by https://github.com/WDscholia/scholia/pull/1702

PR #1618, opened 2021-08-19
Fix table of contents bug

carlinmack:
Previously, the sublist would only be added if there was a following H2. This went unnoticed as this is almost always the case. The affected aspects are event_series.html, event.html, publisher.html, topic.html and works.hml

image


carlinmack:
Perhaps cheeky to add a "ready-to-merge" label to my own PR, but this is very simple and can't really have side-effects

egonw (approved):
Looks good to me. Tested here local and works as reported to me.

PR #1623, opened 2021-08-22
Improve print styles

carlinmack:
Working towards fixing #1031

Add styles to improve printing. I wanted to do a bit more (fix the iframe display, add the link URLs) but couldn't overcome problems in Chrome. There were problems with width in Chrome where the iframes are too narrow and the tables are too wide. 

I also wanted to fix/ease the issue of the WDQS query link appearing in the foreground of the graphs. However, nothing can be done about this as it is in an iframe.

In the end, I've mainly removed the interactive elements (aspect chooser and search) and improved the page margins.

Firefox before and after
image
image

Chrome before and after
image
    
### Caveats

### Testing

* I mainly tested /use/Q45340488 as this has both datatables and iframes, with only a couple of each. 

### Checklist

### DIFF
--- a/scholia/app/static/css/scholia.css
+++ b/scholia/app/static/css/scholia.css
@@ -107,3 +107,31 @@ h3 {
     margin: 1rem;
 
 }
+
+@page {
+    margin: 2cm;
+}
+
+@media print {
+    .container {
+        margin: 0;
+        padding: 0;
+    }
+
+    .table-of-contents a,
+    a.hlink {
+        text-decoration: none;
+        color: black;
+    } 
+
+    .dataTables_filter,
+    #aspect-chooser {
+        display: none;
+    }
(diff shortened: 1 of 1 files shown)


egonw:
Fairly sure I reloaded all resources, a couple of time, but it seems that the margin for the print is not working well in Chrome/Brave (Brave: 1.28.106 Chromium: 92.0.4515.159 (Official Build) (64-bit)):

image

(not a blocker, but likely something you want to fix)

carlinmack:
It's very disheartening how inconsistent print behaviour is across browsers. I'll look into this!

egonw:
@carlinmack, yeah, understood! browser compatibility has given people headaches around the world for 25 years now :) So, please don't consider this a complaint.

Issue #903, opened 2019-10-28
Improve mobile support


Daniel-Mietchen:
navigation, displays, menu, responsive design etc.

PR #1625, opened 2021-08-23
Improving mobile support

carlinmack:
Working towards closing #903

    
First commit in this PR refactors the navbar to be more standard bootstrap, so that it will responsive for mobile 

image

The second commit adds support for mobile 

image

When the menu is expanded
image

The third commit adds horizontal scrolling for datatables, which is implemented with JS. The CSS is needed so that the tables display with the full width on desktop

image

### Caveats

### Testing

* Checked how the menu displayed at iPhone viewport width and 4k width

### Checklist
     * I assume that Bootstrap's defaults are accessible 

### DIFF
--- a/scholia/app/static/css/scholia.css
+++ b/scholia/app/static/css/scholia.css
@@ -57,27 +57,6 @@ h3 {
     box-shadow: 2px 2px;
 }
 
-.navbar {
-    height: 4rem;
-    flex-wrap: nowrap;
-    justify-content: flex-start;
-}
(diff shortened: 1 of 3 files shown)


carlinmack:
Instead of horizontal scrolling on mobile, we could add the Responsive datatables module https://datatables.net/extensions/responsive/examples/styling/compact.html which would instead hide the columns wider than the viewport and they can be expanded below as so: 

image

PR #1628, opened 2021-08-25
Use aspect in link 1192

### LINKED ISSUE #1192: "Use" column in use aspect table should go to use aspects
Items in "Most used" at https://scholia.toolforge.org/use/ typically go to topic aspect. They should go to "use" aspect

DanielFLopez:
fix #1192 

add support for row with Url

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,6 @@
 name: Python package
 
-on: [push]
+on: [push, workflow_dispatch]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


Daniel-Mietchen (changes requested):
I do not think this does what is intended by #1192, so I just added more documentation to the ticket.

fnielsen:
The issue has been fixed with https://github.com/WDscholia/scholia/pull/1765

In is not entirely whether the https://github.com/WDscholia/scholia/pull/1628/commits/b20787cc32f5b67289cf5c4085918ac86494e305 addresses the issue or another issue. AFAIS the 'Url' suffix is already handled.

PR #1635, opened 2021-08-28
Replace select2 with datalist

### LINKED ISSUE #1636: Fix the search bar to not require double clicks
Currently, two clicks &mdash; in different locations &mdash; are required before one can start typing. This is not user friendly.

carlinmack:
Close #1636

Not fully working yet

image

Using a fair bit of code (that I wrote) from my other project https://github.com/carlinmack/afp/blob/main/src/afp-devel/admin/hugo/themes/afp/static/js/search.js#L395

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -501,55 +501,105 @@ sparqlToShortInchiKey(`# tool: scholia
      });
   {% endif %}
 
- $(document).ready(function () {
-     let search_text = ""
-     $('#searchterm').select2({
-         allowClear: true,
-         {% if request.path.endswith("search") %}
-         placeholder: "Scientist, paper, topic, DOI, etc.",
-         {% else %}
-         placeholder: "Search...",
-         {% endif %}
-         ajax: {
-             url: "https://www.wikidata.org/w/api.php?callback=?",
-             dataType: 'jsonp',
-             data: function (params) {
-                 search_text = params.term
-                 var query = {
-                     search: params.term,
-                     action: "wbsearchentities",
-                     language: "en",
-                     uselang: "en",
-                     format: "json",
-                     strictlanguage: true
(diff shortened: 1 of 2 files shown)


Daniel-Mietchen:
I see two issues with this:
- Using this branch, the Wikidata visualizations do not load for me
- autosuggestions/ autocomplete would be useful

See example screenshot for Q211970 :
Screenshot 2021-08-29 at 11-48-47 Scholia

By way of comparison, here is how https://scholia.toolforge.org/location/Q211970 looks to me right now:
Screenshot 2021-08-29 at 11-51-50 Scholia

That the second visualization is missing here too is another issue, though I do not have the ticket at hand.

carlinmack:
Sorry I stated in the post that it's not fully working yet, I haven't done anything with the search boxes on other pages yet. 

In the future it would be better to test PRs that aren't "Drafts"

carlinmack:
Alright after some thorough investigation, I don't think that the datalist approach will work. Two things at play:

1. We want to have users search for a string and retrieve a QID from Wikidata. Unfortunately you cannot use an ID (like in the select2 approach) and you have to use a workaround with text and value which looks like: `<option value="Q55233080">Florence Page Jaques - American author (1890-1972)</option>` and look like this:

image
 or this on Firefox:

image

The UX of this isn't great with the inconsistency, but also in Chrome you can't swap it so the Q number is the subtitle as you can only access the `value` (the big heading) with JS.

2. Second, the browser actually manually filters the datalist based upon the input. This is news to me, and is unfortunate as therefore any value returned by Wikidata which doesn't exactly match won't be shown. Also "Advanced Search" option will not be shown unless it matches. 

Due to these, a different strategy will have to be used. I will commit the version I was playing with now and then close this. I'll open a different PR with the new strategy

Issue #1638, opened 2021-08-29
work metadata can show the language of the work


egonw:
... making it easier to see the difference between these two articles:

![image](https://user-images.githubusercontent.com/26721/131250604-05ed816a-c93a-446c-b1be-7af9aaf072a3.png)

fnielsen:
I am not sure why this wasn't closed automagically.

PR #1639, opened 2021-08-29
Added the language of a work

egonw:
Implements https://github.com/WDscholia/scholia/issues/1638

New table looks like:

image

### DIFF
--- a/scholia/app/templates/work_data.sparql
+++ b/scholia/app/templates/work_data.sparql
@@ -42,6 +42,15 @@ WHERE {
   }
   UNION
   {
+    BIND(2.5 AS ?order)
+    BIND("Language" AS ?description)
+    ?work wdt:P407 ?iri .
+    ?iri rdfs:label ?value_string .
+    FILTER (LANG(?value_string) = 'en')
+    BIND(STR(?value_string) AS ?value)
+  }
+  UNION
+  {
     SELECT
       (3 AS ?order)
       ("Reviewers" AS ?description)


scholia/app/templates/work_data.sparql
```diff
@@ -41,6 +41,15 @@ WHERE {
     GROUP BY ?dummy
   }
   UNION
+  {
+    BIND(2.5 AS ?order)
+    BIND("Language" AS ?description)
+    ?work wdt:P407 ?iri .
+    ?iri rdfs:label ?value_string .
+    FILTER (LANG(?value_string) = 'en')
```
carlinmack, line 49: This will only show for English works?

carlinmack: oh nevermind, that the label itself is in English

carlinmack (approved):

carlinmack:
If you change the first line of the post to ["Fix #" or "Close #" it will "link" the issue and close it when this PR is merged](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)

scholia/app/templates/work_data.sparql
egonw: oh, but a good call, nevertheless! this should be taken from the active UI... hang on, will fix this!

egonw:
> If you change "Implement" in the first line of the post to ["Fix", "Close", etc](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) it will "link" the issue and close it when this PR is merged

yeah, I know. I learned this trick from Debian a very long time ago... but it no longer make sense to have this in the git history, which I reliazed after CVS->Subversion->Git->GitHub ... so, I stopped doing this, sort of intentionally.

scholia/app/templates/work_data.sparql
egonw: Actually, the full query has this... I'll file an issue for now to check all `FILTER` in all SPARQL instead, if you're okay.

egonw: Filed issue: https://github.com/WDscholia/scholia/issues/1640

PR #1641, opened 2021-08-30
Fix item in home page list

carlinmack:
All the other items in the list are plural

Currently 

image


Daniel-Mietchen (approved):
Straightforward.

PR #1652, opened 2021-08-31
Lazy load iframes

jdcaballerov:
This will improve load times and prevent rate limiting.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,6 @@
 name: Python package
 
-on: [push]
+on: [push, workflow_dispatch]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -1,6 +1,6 @@
 name: Python package
 
-on: [push]
+on: [push, workflow_dispatch]
```
egonw, line 3: This is not about the iframes, but sounds reasonable to me.

scholia/app/static/scholia.js
```diff
@@ -261,6 +261,7 @@ function sparqlToIframe(sparql, element, filename) {
     let $iframe = $(element)
     url = "https://query.wikidata.org/embed.html#" + encodeURIComponent(sparql);
     $iframe.attr('src', url);
+    $iframe.attr('loading', 'lazy');
```
egonw, line 264: Sounds useful.

egonw (approved):

fnielsen:
I did try on a Ubuntu Firefox and Chrome and am not sure that it actually works.

PR #1657, opened 2021-09-01
Add link to reasonator and sqid 286

### LINKED ISSUE #286: From Topic aspects, link to Reasonator and SQID
i.e. from
https://tools.wmflabs.org/scholia/topic/Q202864
to
https://tools.wmflabs.org/reasonator/?q=Q202864
and 
https://tools.wmflabs.org/sqid/#/view?id=Q202864

DanielFLopez:
fix #286

add link to reasonator and sqid in topic aspect

Screen Shot 2021-09-01 at 11 34 18 AM


Daniel-Mietchen (approved):
Code looks good and works as expected.

Here is  for the current situation:

And here it is with this PR:


The positioning may merit some further thought &mdash; e.g. we could move it under the "The Topic in context" heading or integrate it with the Wembedder display. 

It is also worth considering to set this up generically across aspects, and in that case, keeping it above the Table of Contents does probably make sense. 

I just opened #1695 for that, so this current PR is good to go in my view.

PR #1660, opened 2021-09-03
Add link to about 879

### LINKED ISSUE #879: Make About link visible from the menu bar directly
rather than through the dropdown. This will affect the "Help" and "FAQ" links as well, so let's consider them in context.

DanielFLopez:
fix #879 

add link visible to about and faq

Screen Shot 2021-09-03 at 4 35 21 PM

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,6 @@
 name: Python package
 
-on: [push]
+on: [push, workflow_dispatch]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


carlinmack (changes requested):
Your PR fixes the issue, but creates another. Why do we now have a dropdown called Help with one item, Statistics?

DanielFLopez:
> Your PR fixes the issue, but creates another. Why do we now have a dropdown called Help with one item, Statistics?

should I remove the help? and leave statistics?

carlinmack (approved):

Daniel-Mietchen (approved):
Looks good.

fnielsen:
I have added a comment on the issue page https://github.com/WDscholia/scholia/issues/879#issuecomment-963289211 for why I do not see this is a good idea.

PR #1684, opened 2021-10-25
CiTO updates

egonw:


Daniel-Mietchen (approved):
Looks good to me.

Here is the situation for https://scholia.toolforge.org/cito/ as of now:
Screenshot from 2021-11-01 20-26-52


And here it is with this PR merged:
Screenshot from 2021-11-01 20-28-00

PR #1707, opened 2021-11-21
Add author score to the missing-author-items query

guyfawcus:
### Description
This adds the author score (using same scoring logic as in [author-scores.sparql](https://github.com/WDscholia/scholia/blob/0f891e9c6be2edad3de0e5b2ce32c4b5b6558afa/scholia/app/templates/topic_author-scores.sparql)) to the missing-author-items query on the topic curation page.

Prior to this, the only way to sort the list of missing authors was by their work count.

After you disambiguate the top authors, you're usually left with a long tail of authors who only have one or two works on the subject. Adding a "score" column to the table to makes it easier to prioritise disambiguation efforts.
    
### Caveats

* This is **not** a breaking change
* It only changes a single SPARQL query
* It should slightly increase the query time but the change is so small that it's hard to measure
* I don't believe any docs need to be changed as I couldn't find anything related to scoring

### Improvements

* I had considered just returning the number of citations but I thought that it'd be better to keep the consistency with the usual scoring logic
* The query could probably be improved but it illustrates what I was after and has already proved useful
* I'm open to any ideas!

### Testing
* I've tested the changes by using the query service directly and by running the Scholia application locally

### Checklist

### DIFF
--- a/scholia/app/templates/topic-curation_missing-author-items.sparql
+++ b/scholia/app/templates/topic-curation_missing-author-items.sparql
@@ -2,6 +2,9 @@ SELECT
   # Number of works with the author
   ?count
 
+  # Author score - 3 points for authoring a work, 1 for being cited
+  ?score
+
   # Author as a string
   ?author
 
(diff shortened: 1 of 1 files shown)


Daniel-Mietchen (approved):
This looks good to me.
Here is an example for Zika virus, https://scholia.toolforge.org/topic/Q202864/curation :
Screenshot from 2021-11-28 03-31-10

fnielsen:
Example: http://127.0.0.1:8100/topic/Q202864/curation

guyfawcus:
Nice! Thanks @Daniel-Mietchen and @fnielsen for the screenshot and link, I'll be sure to add them next time I submit a PR :+1:
That query did a really good job at highlighting how useful this can be.

PR #1629, opened 2021-08-26
Encode doi error no ascii characters 1380

DanielFLopez:
fix #1380 

add a function to remove special characters from URL encoding, also remove any non-ASCII character and blank spaces

fix test_get_metadata

PR #1722, opened 2021-12-13
Encode doi error no ascii characters 1380

egonw:
Rebased PR of https://github.com/WDscholia/scholia/pull/1629


egonw:
@fnielsen, those sounds like very minor issues. Why don't you just fix those? Or do you want to ask me to fix those? Or should Monadical fix their PR?

PR #1752, opened 2022-01-16
Added warning if the article cites an retracted article

egonw:
Warns users if an article cites an retracted article


fnielsen:
Example: http://127.0.0.1:8100/work/Q21145873 (for this particular one, there seems to be multiple entries)

egonw:
Thanks, Finn!

Issue #1807, opened 2022-02-08
Scholia hackathon on 14 and 16 February 2022


Daniel-Mietchen:
We will run a hackathon next week, primarily to reduce the backlog of [open pull requests](https://github.com/WDscholia/scholia/pulls) but also &mdash; particularly on the second day &mdash; to work on community matters.

For details, see https://www.wikidata.org/wiki/Wikidata:WikiProject_Scholia/February_2022_hackathon .

While anyone is of course welcome to participate, we won't have much capacity to assist participants unfamiliar with the Scholia code base this time, so we are not advertising the event much through wiki channels.

We do plan to run similar events in the future where we can provide such assistance.

Daniel-Mietchen:
Besides these pull requests, we also worked on a good number of tickets &mdash; see the screenshots below of the recently updated open and closed issues (the hackathon was ca. 5 and 7 days ago).

![Screenshot 2022-02-22 at 01-26-10 Issues · WDscholia scholia](https://user-images.githubusercontent.com/465923/155042004-0724aad0-69aa-4952-ac3c-c3101f99c297.png)
![Screenshot 2022-02-22 at 01-26-18 Issues · WDscholia scholia](https://user-images.githubusercontent.com/465923/155042001-50243581-6941-4e5e-8608-801c5c5a356c.png)
![Screenshot 2022-02-22 at 01-27-22 Issues · WDscholia scholia](https://user-images.githubusercontent.com/465923/155041999-96874446-25c0-435c-a4dd-f12b3c154e42.png)

Closing.

PR #1843, opened 2022-02-16
Add LIMIT to topic_authors.sparql

### LINKED ISSUE #1842: In Topic aspect, add LIMIT to ```Authors publishing about the topic``` panel
**What kind of panel are you referring to?**
The  ```Authors publishing about the topic``` panel, e.g. as per https://scholia.toolforge.org/topic/Q210861#authors .

**What kind of information should the panel provide, and which of the visualization options (e.g. table, bubble chart, map) should it use?**

It works fine in most cases, just when there is too much information, it might fail, so adding a LIMIT would be helpful.

Screenshot from 2022-02-16 13-52-29

**Which Wikidata entries would be good candidates to explore this?**

Heavily curated topics, such as the ones on [this list](https://w.wiki/4qhS)

Daniel-Mietchen:
Fixes #1842 

Just added a LIMIT to the SPARQL query.


fnielsen:
Shouldn't the `LIMIT` be before the labeling, to avoid spending time on labeling items that are not shown?

Daniel-Mietchen:
I tested this for a number of the highly curated topics, e.g. [schizophrenia](https://scholia.toolforge.org/topic/Q41112). While the LIMIT I chose is 200 (as in many other panels), the number of results we get is sometimes more than that (e.g. 203 right now for Schizophrenia), due to people having multiple ORCIDs. I think this should be fine, but we could make it always 200 by adding ```LIMIT 200``` at the bottom again.

Screenshot from 2022-02-16 14-28-10

scholia/app/templates/topic_authors.sparql
```diff
@@ -1,13 +1,16 @@
+# tool: scholia
 #defaultView:Table
 SELECT ?count ?author ?authorLabel (COALESCE(?orcid_, CONCAT("orcid-search/quick-search/?searchQuery=", ?authorLabel)) AS ?orcid) 
 WITH {
-  SELECT (count(?work) as ?count) ?author WHERE {
+  SELECT DISTINCT ?author (count(?work) as ?count) WHERE {
```
fnielsen, line 5: Is the DISTINCT necessary when we have the GROUP BY?

Daniel-Mietchen:
Adding a pointer to 
- #1807 

which serves to keep an overview of things we worked on during the hackathon.

PR #1876, opened 2022-02-21
Gitpod configuration 1801

### LINKED ISSUE #1801: Gitpod lauch does not work: flask command not found
**Describe the bug**
Gitpod lauch does not work: flask command not found

**To Reproduce**
1. Go to https://gitpod.io/#https://github.com/WDscholia/scholia
2. Error message: "bash: flask: command not found

**Expected behavior**

**Screenshots**
Flask should start in the Gitpod Window.

**Additional context**
Perhaps `python -m flask ...`?

Daniel-Mietchen:
Fixes #1801 

I modified the Gitpod configuration, and it now works as I would expect, giving the result below upon launching 
https://gitpod.io/#https://github.com/WDscholia/scholia/tree/gitpod-configuration-1801 .

Screenshot from 2022-02-21 21-06-19


Daniel-Mietchen:
As this PR removed the lines
```
image:
  file: .gitpod.Dockerfile
```

I also tested the effect of having them in or out &mdash; see the two commits before this comment. With them, I got an error, so I kept them out. This might break something in the Docker workflow, which I have not looked into.

Daniel-Mietchen:
Pinging #1570 regarding the Docker part.

PR #1941, opened 2022-04-10
Update to citation.js 0.5.6 and software-formats 0.5.3

larsgw:
### Description
Update the Citation.js bundle to citation-js 0.5.6 and [plugin-software-formats](https://github.com/citation-js/plugin-software-formats) 0.5.3. See https://github.com/citation-js/plugin-software-formats/issues/12. Bundle content as specified here: https://github.com/WDscholia/scholia/pull/1454

### Checklist

### DIFF
--- a/scholia/app/static/js/citation.js
+++ b/scholia/app/static/js/citation.js
@@ -6,7 +6,7 @@ Object.defineProperty(exports, "__esModule", {
 });
 exports.default = void 0;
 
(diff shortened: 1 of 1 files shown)


larsgw:
The remaining/new CodeQL result is a false positive, it (correctly) sees that the code does not sanitize HTML properly but that is not the purpose, it's to create a valid BibTeX label.

PR #1961, opened 2022-05-16
Org map - Use the headquarters location by default

guyfawcus:
It seems as though the preferred method of storing an organisation's location is by adding a [P625](http://www.wikidata.org/entity/P625) (coordinate location) qualifier on a [P159](http://www.wikidata.org/entity/P159) (headquarters location) statement.

This commit changes the org map query to use the headquarters location by default but still keeps the coordinate location as a fallback.

Relevant discussions:
* https://www.wikidata.org/wiki/Wikidata:Project_chat/Archive/2021/03#coordinate_location_(P625)_to_headquarters_location_(P159)_switch
* https://www.wikidata.org/wiki/Wikidata:Project_chat/Archive/2022/04#DeltaBot's_coordinate_location_(P625)_to_headquarters_location_(P159)_move

### DIFF
--- a/scholia/app/templates/topic_organization-map.sparql
+++ b/scholia/app/templates/topic_organization-map.sparql
@@ -16,7 +16,11 @@ WITH {
     # Authors who have published works on the topic
     ?work wdt:P50 ?author . 
     ?author ( wdt:P108 | wdt:P463 | wdt:P1416 ) / wdt:P361* ?organization . 
-    ?organization wdt:P625 ?geo .
+    # Use the headquarters location by default but keep the coordinate location as a fallback
+    OPTIONAL{?organization p:P159/pq:P625 ?hq_geo}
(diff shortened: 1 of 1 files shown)


fnielsen:
Do we have example that can show the solution?

Daniel-Mietchen:
@fnielsen @guyfawcus here is an example: https://scholia.toolforge.org/topic/Q6900886#organization-map

Before:
Screenshot from 2022-05-21 16-34-20


After (with one difference highlighted):
Screenshot from 2022-05-21 16-33-53

guyfawcus:
Oops, [thanks again](https://github.com/WDscholia/scholia/pull/1707#issuecomment-981748152) @Daniel-Mietchen for adding some screenshots! I'll try harder next time :sweat_smile:

PR #1965, opened 2022-05-18
fix the list-of-works heading

### LINKED ISSUE #1964: Fix the "List of works on with description of uses" heading in uses.html
**Describe the bug**
"List of works on with description of uses" is not proper English and should be replaced by something like "List of works with descriptions of uses"

Daniel-Mietchen:
Fixes #1964


fnielsen:
Now running: https://scholia.toolforge.org/uses/Q2013,Q52

PR #1978, opened 2022-05-21
adding case variants to award-curation_missing-author-items.sparql

### LINKED ISSUE #1977: Modify award-curation_missing-author-items.sparql to include case variants of author name strings
**What query is this about**
The way author name strings are capitalized can vary quite a bit even for the same author (cf. #1854 ). This should also be reflected in award-curation_missing-author-items.sparql too.

**What change do you propose, and why?**

Include case variants as in https://github.com/WDscholia/scholia/pull/1976 .

**Anything else to consider here?**

There are similar queries on other curation pages &mdash; will try to go through those too.

Daniel-Mietchen:
Fixes #1977 

### Description

added case variants.
Example: https://scholia.toolforge.org/award/Q38104/curation#missing-author-items

#### Before
Screenshot from 2022-05-21 23-09-48
Screenshot from 2022-05-21 23-10-32

#### After

Note that the results differ in part due to the ```LIMIT```, in part due to this change.

Screenshot from 2022-05-21 23-10-51
Screenshot from 2022-05-21 23-11-29

### Checklist

### DIFF
--- a/scholia/app/templates/award-curation_missing-author-items.sparql
+++ b/scholia/app/templates/award-curation_missing-author-items.sparql
@@ -1,27 +1,53 @@
 PREFIX target: <http://www.wikidata.org/entity/{{ q }}>
 
 SELECT
-  (COUNT(?work) AS ?count) 
-  ?author_name
-  (CONCAT(
-      'https://author-disambiguator.toolforge.org/names_oauth.php?doit=Look+for+author&name=',
-      ENCODE_FOR_URI(?author_name)) AS ?author_resolver_url)
-WHERE {
-  {
-    SELECT DISTINCT ?author_name {
-      ?recipient p:P166 ?award_statement . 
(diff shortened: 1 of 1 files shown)


scholia/app/templates/award-curation_missing-author-items.sparql
```diff
@@ -1,27 +1,53 @@
...
+    ?award_statement ps:P166 target: .
+
+    ?recipient skos:altLabel | rdfs:label ?author_name_ .
+
+    # The SELECT-DISTINCT-BIND trick here is due to Stanislav Kralin
+    # https://stackoverflow.com/questions/53933564
+    BIND (STR(?author_name_) AS ?author_name)
+  }
+  LIMIT 2000
+} AS %RAWstrings
```
fnielsen, line 21: Maybe we should stick to lowercase variable names.

PR #2021, opened 2022-06-14
Link venue in venue column

### LINKED ISSUE #1481: On author profiles under Venue statistics, the venue column in the table should link to the venue aspect
Example from https://scholia.toolforge.org/author/Q88802362#venue-statistics-header :

Screen Shot 2021-07-05 at 06 48 23

The "AIDS and Behavior" link there currently goes to https://scholia.toolforge.org/Q15752229 , which redirects to https://scholia.toolforge.org/topic/Q15752229 , when https://scholia.toolforge.org/venue/Q15752229 would be more useful.

carlinmack:
Fixes #1481

Generally adds the ability for a column to have a label and a URL. Previously they could have only have one.

Specifically link the venue column to the venue aspect

image

### Caveats

### Testing

* Checked on the example page. Should only affect queries where there is a label and a url

### Checklist

### DIFF
--- a/scholia/app/static/scholia.js
+++ b/scholia/app/static/scholia.js
@@ -103,43 +103,54 @@ function convertDataTableData(data, columns, linkPrefixes = {}, linkSuffixes = {
 
 	    } else if (
 			key + 'Label' in data[i] &&
+            key + 'Url' in data[i]
+	    ) {
+            convertedRow[key] = '<a href="' +
+		    data[i][key + 'Url'] +
+		    '">' + data[i][key + 'Label'] + '</a>';
+
+	    } else if (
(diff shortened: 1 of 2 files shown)


fnielsen:
There is optional arguments in the Javascript to redirect to other aspects, so we have two methods for doing it. Perhaps the "in SPARQL" redirect is better than the "in-Jinja/javascript" redirect?

carlinmack:
Yeah I think the in-SPARQL redirect is better as it's easier to customise and the mechanism is more visible

carlinmack:
I can follow up with a PR to refactor the other queries to use in-SPARQL :)

PR #2025, opened 2022-06-15
Remove linkPrefixes option and replace with SPARQL

carlinmack:
This builds on #2021

Fixes # (issue number, if applicable) 

    
Removes linkPrefixes in favour of using SPARQL. There are a couple reasons to do this:

* it's more natural for someone contributing a new aspect to write only SPARQL rather than also specifying the JS
* With linkPrefixes the developer needs to be aware of the number of pages deep the aspect is (`linkPrefix{ "author": "../../../author"}`), which is brittle to restructuring
* it's less flexible than defining it in SPARQL
* there are a few cases where the linkPrefixing is wrong and I think it would probably be more obvious if it was in the SPARQL. When you're making the query you can check the Url looks right whereas with JS you need to render the aspect on the page itself to check

I have split my commits into the different cases that appeared when doing this. Happy to go into detail about any of them

### Caveats

### Testing

* I have checked <10 pages, but I have looked over the changes in this PR to ensure they look okay

### Checklist

### DIFF
--- a/scholia/app/static/scholia.js
+++ b/scholia/app/static/scholia.js
@@ -76,7 +76,7 @@ function scroll() {
 /* scroll to anchor in case page has changed */
 document.addEventListener('DOMContentLoaded', () => {setTimeout(scroll, 2000)});
 
-function convertDataTableData(data, columns, linkSuffixes = {}) {
+function convertDataTableData(data, columns) {
     // Handle 'Label' columns.
 
     var convertedData = [];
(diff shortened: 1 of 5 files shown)


fnielsen:
There is currently a conflict

fnielsen:
There are remaining linkPrefixes and the 'linkSuffixes' is also a problem

carlinmack:
All done

PR #2080, opened 2022-08-08
Fix #2079 Too long doc lines

fnielsen:
Fixes #2079

### Description
Fixes styling errors
    
### Caveats
There are still some test errors in tox for python3.8 and python3.9

* [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ]  This change requires a documentation update
    * [ ]  I have made corresponding changes to the documentation
* [ ]  This change requires new dependencies (please list)

*If you make changes to the Python code*
  
* [ ]  My code passes the [tox](https://tox.readthedocs.io/en/latest/) check, I can receive warnings about tests, documentation or both

### Testing
`pycodestyle scholia` runs ok. 

`tox py38` does not run ok

### Checklist
* [x] I have [commented](https://peps.python.org/pep-0008/#comments) my code, particularly in hard-to-understand areas
* [ ] My changes generate no new warnings
* [x] I have not used code from external sources without attribution
* [x] I have considered [accessibility](https://www.w3.org/standards/webdesign/accessibility) in my implementation 
* [x] There are no remaining debug statements (print, console.log, ...)

PR #2075, opened 2022-08-04
Limits for use profile 2070

Daniel-Mietchen:
Fixes #2070

Added LIMITs to most of the queries and subqueries on the use aspect
    
### Caveats

The use of LIMITS in the (usually unsorted) subqueries means that once those limits are reached (and some are already), then relevant results will not be included in the visualizations. In such cases, it may be more useful to actually have a random selection via something like 

```SPARQL
SERVICE bd:sample { ?work wdt:P4510 target: . bd:serviceParam bd:sample.limit 100000 }
```

> Please list anything which has been left out of this PR or which should be considered before this PR is accepted
Check any of the following which apply:

* [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ]  This change requires a documentation update
    * [ ]  I have made corresponding changes to the documentation
* [ ]  This change requires new dependencies (please list)

*If you make changes to the Python code*
  
* [ ]  My code passes the [tox](https://tox.readthedocs.io/en/latest/) check, I can receive warnings about tests, documentation or both

### Checklist
* [ ] I have [commented](https://peps.python.org/pep-0008/#comments) my code, particularly in hard-to-understand areas
* [X] My changes generate no new warnings
* [X] I have not used code from external sources without attribution
* [X] I have considered [accessibility](https://www.w3.org/standards/webdesign/accessibility) in my implementation 
* [X] There are no remaining debug statements (print, console.log, ...)

### Screenshots
Example: https://scholia.toolforge.org/use/Q1659584
Screenshot 2022-08-04 at 23-01-22 ImageJ - Scholia

PR #2082, opened 2022-08-09
Update chemical.html

Adafede:
### Caveats

> None

### Testing

> None

### Checklist

### DIFF
--- a/scholia/app/templates/chemical.html
+++ b/scholia/app/templates/chemical.html
@@ -50,7 +50,9 @@
 
 <table class="table table-hover" id="identifiers-table"></table>
 
-<h2 id="Related" id="related">Related Compounds (including the compound itself)</h2>
+<h2 id="Related" id="related">Related Compounds</h2>
+
+(Including the compound itself)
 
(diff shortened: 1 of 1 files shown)


Adafede:
Thank you, did not think about that.
Updated accordingly

fnielsen:
I am wondering why this PR has no testing issues while, e.g., https://github.com/WDscholia/scholia/pull/2075 errs

fnielsen:
The "All checks have passed" seems not to run Python tests. I have now run `tox` locally and find no errors for "Adafede-patch-3":
```
  flake8: commands succeeded
  pydocstyle: commands succeeded
  py37: commands succeeded
  py38: commands succeeded
  py39: commands succeeded
  congratulations :)
```
Yet https://github.com/WDscholia/scholia/pull/2080 which is based on the same master has errors in tox testing. I do not understand how this is possible.

Adafede:
I was actually worried about my [branch](https://github.com/Adafede/scholia/runs/7741187731?check_suite_focus=true) because of the errors also but got it working here with the PR...

PR #2097, opened 2022-09-06
Add message when query returns no matches

### LINKED ISSUE #2037: Javascript error in table formatting
**Describe the bug**
```
Uncaught TypeError: convertedData.data[0] is undefined
    sparqlToDataTable https://scholia.toolforge.org/static/scholia.js:304
```

**To Reproduce**
1. Go to https://scholia.toolforge.org/author/Q18618629
2. See error in Javascript console.

carlinmack:
Fixes #2037

    
This was a great bug to catch! As of now when a query with a count column returns no matches it errors and loads forever:
image

I have instead added a message to communicate to the user

image

Additionally I have removed the loader when the query fails

### Caveats

### Testing

* Test A
* Test B

### Checklist

### DIFF
--- a/scholia/app/static/scholia.js
+++ b/scholia/app/static/scholia.js
@@ -285,53 +285,75 @@ function sparqlToDataTable(sparql, element, filename, options = {}) {
 
         convertedData = convertDataTableData(simpleData.data, simpleData.columns);
         columns = [];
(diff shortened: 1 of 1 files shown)


carlinmack:
I've just added an `if` around existing code but the diff looks so much more complicated than that, apologies

PR #2105, opened 2022-09-11
Add refreshing of data tables

### LINKED ISSUE #886: Add a refresh button to each panel
That would
- be achievable by linking that button to a query that differs from the one being displayed by just a # or space or so at the end
- help with curation

carlinmack:
Will fix the issue #886 partially

    
To do:
- [x] Refreshing of Datatables
- [ ] Refreshing of iFrames
- [x] Styling of refresh button
- [x] UX in regard to highlighting the button if the query times out

Result: 

image

The loader briefly appears and the page looks the same when the reload button is clicked

When there is no header it appears just before it in the HTML

image

Without `clear: both` on `H3` we have:
image
image

I've added it so instead we have
image

### Caveats

### Testing

* Checked a broad variety of pages and made sure refreshing works. Unfortunately haven't been able to test a case where the query fails and reloading helps as it's hard to find 

### Checklist

### DIFF
--- a/scholia/app/static/css/scholia.css
+++ b/scholia/app/static/css/scholia.css
@@ -152,6 +152,7 @@ h2 {
 
 h3 {
     font-size: 1.65rem;
+    clear: both;
 }
 
 .dropdown-menu li {
--- a/scholia/app/static/scholia.js
+++ b/scholia/app/static/scholia.js
@@ -234,7 +234,7 @@ function sparqlToDataTablePost(sparql, element, filename, options = {}) {
           paging = false;
         }
 
-        $(element).html('');
+        $(element).html(''); // remove loader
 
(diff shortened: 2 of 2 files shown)


carlinmack:
Using https://commons.wikimedia.org/wiki/File:Refresh_icon.svg for the button icon

carlinmack:
Refreshing iframes seems to be extraordinarily difficult :( I think perhaps for security reasons? Will mark this as ready for review

fnielsen:
Could you rebase this. I am getting a problem with merging.

fnielsen:
> Are there any other issues you want me to look at btw?

A recent PR made an error in our Javascript apparent: https://github.com/WDscholia/scholia/issues/2114

Issue #1945, opened 2022-04-19
Fix code quality as reported on lgtm


fnielsen:
**Describe the bug**
https://lgtm.com/projects/g/WDscholia/scholia?mode=list reports poor code quality on various parts of Scholia. 

This is both in the Python and the Javascript code.

carlinmack:
Quite a few of these look serious!

carlinmack:
LGTM is closing in December https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/

I'll look into this

carlinmack:
![image](https://user-images.githubusercontent.com/6676843/195083142-51aae2da-54f0-45c1-b619-8a81ca58f838.png)

I'll make a PR per class of error for ease of reviewing, and list them here too:

- [x] 2 alerts of Flask app is run in debug mode across 2 files — I think this is okay unless we are running in debug mode in production?
- [ ] 1 alert of Suspicious unused loop iteration variable in _version.py
- [ ] 1 alert of Malformed id attribute in author.html
- [ ] 3 alerts of Missing variable declaration in scholia.js
- [ ] 1 alert of Polynomial regular expression used on uncontrolled data in arxiv.py
- [ ] 8 alerts of Unnecessary 'else' clause in loop across 2 files
- [ ] 4 alerts of Module is imported with 'import' and 'import from' across 4 files
- [ ] 3 alerts of Semicolon insertion in scholia.js
- [ ] 2 alerts of Unused variable, import, function or class in scholia.js

carlinmack:
It is easy to fix this error by replacing `i` with `_`:

![image](https://user-images.githubusercontent.com/6676843/195096173-5de037df-bd4c-4d32-ad1e-0f55c576deca.png)

However I think that the for loop could be replaced with its body

https://github.com/WDscholia/scholia/blob/master/scholia/_version.py#L498-L499

carlinmack:
I've also read into this and I don't think it's a problem for us

![image](https://user-images.githubusercontent.com/6676843/195666485-0a6602ce-3a82-44f2-8f53-8b3cd1edf06e.png)

Perhaps we could use `search` instead: "[`findall()` matches all occurrences of a pattern, not just the first one as `search()` does.](https://docs.python.org/3/library/re.html#finding-all-adverbs)" but I don't think it's that important

carlinmack:
All issues have been been considered :) See above

carlinmack:
[...]

PR #2129, opened 2022-10-13
Remove unnecessary else in loop

carlinmack:
For #1945

While this is valid code and is clear, it could be confused that the indentation is incorrect as the syntax is rarely used. As such it is better to remove them as it is equivalent https://lgtm.com/rules/5980098/
    
### Caveats

### Testing

* Test A
* Test B

### Checklist

### DIFF
--- a/scholia/api.py
+++ b/scholia/api.py
@@ -65,12 +65,12 @@ def entity_to_smiles(entity):
     for statement in entity['claims'].get('P2017', []):
         smiles = statement['mainsnak']['datavalue']['value']
         return smiles
-    else:
-        for statement in entity['claims'].get('P233', []):
-            smiles = statement['mainsnak']['datavalue']['value']
-            return smiles
-        else:
-            return ''
+
+    for statement in entity['claims'].get('P233', []):
+        smiles = statement['mainsnak']['datavalue']['value']
+        return smiles
+    
+    return ''
 
 
 def is_human(entity):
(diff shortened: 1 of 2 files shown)


scholia/api.py
```diff
@@ -284,8 +284,8 @@ def entity_to_doi(entity):
     for statement in entity['claims'].get('P356', []):
         doi = statement['mainsnak']['datavalue']['value']
         return doi
-    else:
-        return ''
+
+    return ''
```
fnielsen, line 288: The documentation to the function says that the function returns an empty string if the field is not set.

scholia/api.py
```diff
@@ -313,8 +313,8 @@ def entity_to_full_text_url(entity):
     for statement in entity['claims'].get('P953', []):
         url = statement['mainsnak']['datavalue']['value']
         return url
-    else:
-        return ''
+
+    return ''
```
fnielsen, line 317: The documentation to the function says that the function returns an empty string if the field is not set.

scholia/api.py
```diff
@@ -457,8 +457,8 @@ def entity_to_pages(entity):
     for statement in entity['claims'].get('P304', []):
         pages = statement['mainsnak']['datavalue']['value']
         return pages
-    else:
-        return ''
+
+    return ''
```
fnielsen, line 461: The documentation to the function says that the function returns an empty string if the field is not set.

scholia/api.py
```diff
@@ -505,8 +505,8 @@ def entity_to_volume(entity):
     for statement in entity['claims'].get('P478', []):
         volume = statement['mainsnak']['datavalue']['value']
         return volume
-    else:
-        return ''
+
+    return ''
```
fnielsen, line 509: The documentation to the function says that the function returns an empty string if the field is not set.

PR #2149, opened 2022-10-23
Link works to their (in Wikidata existing) preprints in the data panel

### LINKED ISSUE #2147: Articles do not link to their preprints (when given in Wikidata)
**Is your feature request related to a problem? Please describe.**
Consider you found an article in Wikidata, but the paper is not Open Access. Then, the preprint (often largely the same) can provide you with access to the article.

**Describe the solution you'd like**
Therefore, it would be welcome if the data panel of the `work` aspect would link to the preprint.

I intend to write a patch today.

**Describe alternatives you've considered**
Alternatives, but not very satisfactory, is that the user just finds the preprint via other routes (very hard, since journals refuse to do that).

egonw:
Fixes #2147

I added a row to the data panel that shows a link to the preprint. It uses the pattern `wdt:P31 'preprint'` with the `of` qualifier (in reverse).

The result should now look for `/work/Q114821829` like:

image

### DIFF
--- a/scholia/app/templates/work_data.sparql
+++ b/scholia/app/templates/work_data.sparql
@@ -242,5 +242,17 @@ WHERE {
     BIND(COALESCE(?value_string, ?q) AS ?value)
     BIND(CONCAT("../work/", ?q) AS ?valueUrl)
   }
+  UNION
+  {
+    BIND(32 AS ?order)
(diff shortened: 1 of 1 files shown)


fnielsen:
Running: https://scholia.toolforge.org/work/Q98577324

PR #2154, opened 2022-10-31
Add error handling to fetch and ajax calls

### LINKED ISSUE #1661: All $.ajax() calls should have a defined error case
More specific version of #1026

Right now, all ajax calls have a `.then`/`success:` case and errors are hidden in the console (I'm guessing). This means that if a query fails, the user has no feedback. This lack of feedback is probably why #1659 does not have a definitive error associated with it right now. 

To be formal, we should be aiming to meet criteria 1 and 9 from https://www.nngroup.com/articles/ten-usability-heuristics/ — **for each ajax call failure state, we should be showing a message on the page in clear language.**

A great PR would even aim to implement criteria 5, such that it would retry the call before showing the message or allowing the user to retry the query #886  (but **not** if "Rate limit exceeded" #302). This is tricky.

In regards to reporting errors to the server from the parent issue, I don't think we have the capacity to monitor user error logs at this point.

carlinmack:
Fixes #1661

    
**Before and afters**

1. If summary fails, display enwiki link
image

2. Wikiversity
image

3. If search is unresponsive it will say "Search failed" in the autocomplete rather than showing nothing
4. If ask query fails it will hide the panels rather than showing them
3. Use `console.error` rather than `console.log` for error messages
5. Make sure all ajax calls have a `.fail` or `error:`

### Caveats

### Testing

* Checked error cases except the bioschemas things where I only changed it to console.error

### Checklist

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -433,6 +433,8 @@ askQuery("{{ panel }}", `# tool: scholia
 		 }
 		 $("#wikiversity-extract").text(text).append(html);
 	     }).fail(function(d, textStatus, error) {
+		 var html = "Read on the <a href=\"" + wikiversityUrl + "\">English Wikiversity</a>";
+		 $("#wikiversity-extract").append(html);
 		 console.error("getJSON failed, status: " + textStatus + ", error: "+error)
 	     });
 	 }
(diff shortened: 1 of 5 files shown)


scholia/app/templates/base.html
```diff
@@ -665,7 +676,7 @@
       resolver: 'custom',
       events: {
         search: debounce((searchTerm, callback) => {
-          var url = "https://www.wikidata.org/w/api.php?callback=?";
+          var url = "https://www.wikdata.org/w/api.php?callback=?";
```
fnielsen, line 679: There is a typo here: wikdata -> wikidata.

carlinmack: oh good catch I changed the URLs for testing purposes

PR #2167, opened 2022-11-09
[WIP] Outline of an ontology aspect

### LINKED ISSUE #2165: Create an aspect for ontologies
**What kind of aspect would you like to add to Scholia?**

An aspect about ontologies

**What would be good candidates for visualizations within such an aspect?**

- list of related ontologies, ranked by number of overlapping concepts
- list of publications using the ontology

**Which Wikidata entries would be good candidates to explore such visualizations?**

- https://scholia.toolforge.org/topic/Q135085 (Gene Ontology)

**Which of the existing aspects could or should link to or from this new aspect?**

- topic
- use

**Anything else?**
We should keep related concepts in mind here, e.g. schemas, controlled vocabularies, thesauri and such

egonw:
Fixes #2165 

    

### Testing
Test with the following links:

* http://127.0.0.1:8100/Q55118584 (redirect test)
* http://127.0.0.1:8100/ontology/Q55118584 (aspect about a single ontology)
* http://127.0.0.1:8100/ontology/ (index page)

### Checklist

### DIFF
--- a/scholia/app/templates/ontology-base.html
+++ b/scholia/app/templates/ontology-base.html
@@ -1,0 +1,39 @@
+{% extends "base.html" %}
+
+
+{% block navbar %}
+<div class="navbar navbar-static-top">
+    <ul class="nav nav-pills">
+	<li role="presentation"><a href="..">Scholia</a></li>
+	<li role="presentation"><a href="../author/">Author</a></li>
+	<li role="presentation"><a href="../work/">Work</a></li>
+	<li role="presentation"><a href="../organization/">Organization</a></li>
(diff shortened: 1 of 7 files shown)


fnielsen:
An entry for 'ontology' is missing in base.html, so the 'ontology' aspect is not available in the menu unless one is on the "ontology" page. We have *-base.html files, but why do we have them? Does `class="active"`do anything?

fnielsen:
The *-base.html does not seem to be used: 
```
$ grep "\-base.html" *
base.html:{% extends "bootstrap-base.html" %}
```

PR #2181, opened 2022-11-22
Thought: Do not list people that no longer work at the affiliation

egonw:
@fnielsen, @Daniel-Mietchen, this is just a thought, but instead of filing an issue, I wanted to show some code. Therefore a PR. But it's really only a discussion point / question.

The current `organization` aspect shows everyone who ever has been linked to the affiliation. And since it is sorted by citations count (should we keep that at all?), people who published a lot after they left the institute may give an distorted picture of the organization.

So, I have been playing with the idea of using the `end time` qualifier. This patch shows the use of that.

But there are caveats. For example, this query probably does not handle situation well where someone is *back* at the organization (and currently working there), but since those would be different statements, it might just do it right.

What do you think?

### DIFF
--- a/scholia/app/templates/organization_employees-and-affiliated.sparql
+++ b/scholia/app/templates/organization_employees-and-affiliated.sparql
@@ -13,10 +13,18 @@ WITH {
   } 
 } AS %researchers
 WITH {
+  SELECT DISTINCT ?researcher ?until WHERE {
+    INCLUDE %researchers
+    MINUS { ?researcher p:P108 ?statement . ?statement ps:P108 target: ; pq:P582 ?until . FILTER ( ?until < NOW() ) }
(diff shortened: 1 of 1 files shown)


egonw:
# Without filter (current situation)

image

# With filter (after patch idea)

image

fnielsen:
I haven't made up my mind on this. I suppose some kind of button or an extra column could also be a solution. @egonw Do you want me to merge your PR for now?

egonw:
@fnielsen, not sure either yet. A column has crossed my mind too. @Daniel-Mietchen, what do you think?

Daniel-Mietchen:
I am also of two minds about this and think the current query is a bit more suitable right now given the data - start dates and particularly end dates are rather incomplete and poorly referenced (or even referenceable at all).

We could, however, have multiple queries in the Affiliates section, including this new one.

Another query in this space that I would like to see is organizations that frequently are the previous/ next or even parallel affiliation of the target organization.

PR #2185, opened 2022-11-26
Report venue is CiTO annotation source

egonw:
Right now, this query report the number of articles with CiTO annotation over time and since recently typed by whether or not the article itself provided explicit citation intentions:

image

This patch updates this to not report `in article` but report the venue with the explicit annotations in the article. The point of this is to visualize the uptake of explicit annotation by venues:

image

(I can imagine that in the future when this has become commonplace, we will depict this quite differently again)


fnielsen:
I am not entirely sure I understood this. :)

PR #2214, opened 2022-12-30
Panel for the author curation page to list articles that are not used as reference for any statement

### LINKED ISSUE #2213: only 0.67% of the articles in Wikidata are used to support statements
On the Telegram channel for Wikidata, they looked into how many articles are actually used as reference to support a claim. That turned out to be 263,247. Or, 0.67%.

So, another curation people can do around an author (e.g. by an author) is to use their articles as "citation" in statements.

egonw:
Fixes #2213 

### Description
This patch adds a panel to an author curation page, listing works for that author that do not support any
statements.
    
### Caveats
No caveats I can foresee.

### Testing
Test is with any author with multiple articles. The output should look something like this:

image

### Checklist

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -108,6 +108,7 @@ askQuery("{{ panel }}", `# tool: scholia
 	 if ('en' in item.labels) {
 	     var title = item.labels.en.value
 	     $("#h1").text(title);
+	     $(".self").text(title);
        $("title").text(title + " - Scholia");
 
 	     {% if request.path.endswith("curation") or request.path.endswith("curation/") %}
--- a/scholia/app/templates/author-curation.html
+++ b/scholia/app/templates/author-curation.html
@@ -55,10 +55,10 @@ the authors.
 <h2 id="missing-use-as-reference">Articles not used as statement reference</h2>
 
 <p>Works authored by <span class="self">{{ q }}</span> that are not used as "stated in" reference
-to a statement. You can improve the Wikidata by finding an unsource statement
-in Wikidata that is supported by any of the listed articles by scrolling that
+to a statement. You can improve Wikidata by finding an unsourced statement
+in Wikidata that is supported by any of the listed articles by scrolling to that
 statement and clicking "+ add reference" and adding a <a href="https://www.wikidata.org/wiki/Property:P248">"stated in" (P248)</a>
-with the supporting article as value.</p>
+with the supporting article's Wikidata identifier (QID) as value.</p>
 
 <table class="table table-hover" id="missing-use-as-reference-table"></table>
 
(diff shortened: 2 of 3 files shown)


Daniel-Mietchen:
In the line ```Works authored by <QID> that do not have topics listed in Wikidata```, the QID should probably be converted to the name string (same issue with the missing topics query above).

egonw:
> In the line `Works authored by <QID> that do not have topics listed in Wikidata`, the QID should probably be converted to the name string (same issue with the missing topics query above).

do we have a template solution for, like `{{ q }}` for the identifier?

Daniel-Mietchen:
Not really a template, but here is a way that works: https://github.com/WDscholia/scholia/blob/c0840e86adabfb3f7e4b69e84ee6d9054b1e39d5/scholia/app/templates/base.html#L106 . Slightly different way a bit further down. Should perhaps be turned into a function.

egonw:
Yeah, that requires a feature in the core library. I like to not make that part of this patch. Actually, after looking at the `base.html`, I think our approach needs some rethinking. Injecting curation content in the `base.html` does not look good design to me (and I'm also wondering if we want Bioschemas on the curation pages).

Daniel-Mietchen:
This seems to be a pragmatic solution for now.

I also looked into the CI errors but did not manage to get rid of them.

egonw:
> This seems to be a pragmatic solution for now.

Sure. This PR is not a code clean up patch, but you asked for this. What you ask for a redesign of the core of Scholia. I think we need that, but I don't think that's inside the scope of patches that add panels.

> I also looked into the CI errors but did not manage to get rid of them.

Yeah, they seem to be random and not related to the changes. I saw it for other PRs too and rerunning them tends to help. Will do that here too.

fnielsen:
There are now two issues for testing problems: #2217 and #2218

fnielsen:
I have now fixed #2217. Possibly as rebase would get the checks successful

egonw:
> I had another look at the tox error but did not find any solution.

It seems the rebase solved the tox issues, not?

fnielsen:
I think with the changes I made, the tox seems to be ok now.

scholia/app/templates/base.html
```diff
@@ -108,6 +108,7 @@
 	 if ('en' in item.labels) {
 	     var title = item.labels.en.value
 	     $("#h1").text(title);
+	     $(".self").text(title);
```
fnielsen, line 111: How is that related to the new panel?

Daniel-Mietchen: It addresses the issue of using the item's label in the section's explanatory text, as discussed [here](https://github.com/WDscholia/scholia/pull/2214#issuecomment-1367983965).

PR #2215, opened 2022-12-31
New property for crystal structures, new statistics

egonw:
### Description
Small patch: when calculating the number of crystal structures, compounds with a CSD Refcode can be counted too. This property was accepted this week.
    
### Caveats
Potentially, a more complex makes the query not run fast enough, but this does not seem to be the case
(not noticeable ).

### Testing
Visit https://scholia.toolforge.org/chemical/ (before/after)

### Checklist
* [ x I have not used code from external sources without attribution


fnielsen:
Would we get a double count?

`SELECT (COUNT(*) AS ?count) WHERE { wd:Q37129 wdt:P3636 | wdt:P11375 []. }`

returns 3. Is that what we intend?

PR #2225, opened 2023-01-14
Renamed the "Related Compounds" section into "Compounds with same connectivity"

### LINKED ISSUE #1942: Chemical: "Related Compounds" renaming
## Context
I was browsing some info about https://scholia.toolforge.org/chemical/Q3613679 and noticed for the first time the `Related Compounds` panel.

## Question
- Could we rename it more precisely? The result of the [chemical_related.sparql](https://github.com/WDscholia/scholia/blob/master/scholia/app/templates/chemical_related.sparql) are **known stereoisomers** which is more defined than `Related compounds`.

## Suggestion
In case the initial goal was not only to list *known stereoisomers* but `related compounds`(related, how? structurally?), a possibility would be to use the [Sachem cartridge](https://doi.org/10.1186/s13321-018-0282-y). (example: [https://w.wiki/4VG4](https://w.wiki/4VG4))

@egonw

egonw:
Fixes #1942

### Description
Renames the section as suggested in the issue.
    
### Caveats

None.

### Testing

* check https://scholia.toolforge.org/chemical/Q3613679#Related

### Checklist

### DIFF
--- a/scholia/app/templates/chemical.html
+++ b/scholia/app/templates/chemical.html
@@ -50,7 +50,7 @@
 
 <table class="table table-hover" id="identifiers-table"></table>
 
-<h2 id="Related" id="related">Related Compounds</h2>
+<h2 id="Related" id="related">Compounds with same connectivity</h2>
 
 (Including the compound itself)
 


scholia/app/templates/chemical.html
```diff
@@ -50,7 +50,7 @@ <h2 id="Identifiers" id="identifiers">Identifiers</h2>
 
 <table class="table table-hover" id="identifiers-table"></table>
 
-<h2 id="Related" id="related">Related Compounds</h2>
+<h2 id="Related" id="related">Compounds with same connectivity</h2>
```
Daniel-Mietchen, line 53: Should the id  be changed accordingly to avoid similar confusion later?

egonw: There is a duplication here and some other places. I propose we fix that in a different patch: we need to check all files.

egonw: I am not a big fan overloading patches.

egonw: Just to not forget about it, I filed https://github.com/WDscholia/scholia/issues/2226

Daniel-Mietchen: OK.

Daniel-Mietchen (approved):
With the id issue ported to #2226, I think this one is good to go.

PR #2271, opened 2023-04-01
Added the number of datasets with explicit CiTO annotations

egonw:
New feature.

### Description
Besides journal articles and individual annotations, there are also datasets with explicit CiTO annotation, e.g. as the outcome of a literature study. This addition uses a new Wikidata item for datasets with explicit CiTO annotation, allowing us a way to track them. There are three that I currently know of, two of which are in Wikidata.
    
### Caveats

None that I can think of.

### Testing

* open https://scholia.toolforge.org/cito/ and notice the extra line:

image

### Checklist


fnielsen:
Running at https://scholia.toolforge.org/cito/

Issue #1853, opened 2022-02-16
Add [PID/QID] aspect


Daniel-Mietchen:
**What kind of aspect would you like to add to Scholia?**

I would like to have an aspect that can be used by WikiProjects, e.g. 
https://scholia.toolforge.org/P5008/Q56241615 , where [P5008](https://www.wikidata.org/wiki/Property:P5008) ```is on focus list of Wikimedia project```, and [Q56241615](https://www.wikidata.org/wiki/Q56241615) is WikiProject Invasion Biology.

**What would be good candidates for visualizations within such an aspect?**

- Publications within scope of the WikiProject
- Authors within scope of the WikiProject
- Institutions within scope of the WikiProject
- Taxa within scope of the WikiProject
- Locations within scope of the WikiProject
- Events within scope of the WikiProject
- Journals within scope of the WikiProject

etc. For some more examples, see https://www.wikidata.org/wiki/Wikidata:WikiProject_Invasion_biology

**Which Wikidata entries would be good candidates to explore such visualizations?**

- Any of the ones in https://scholia.toolforge.org/P5008#property-value-counts , especially
  - [WikiProject Zika corpus](https://www.wikidata.org/wiki/Q54439832)
  - [WikiProject COVID-19](https://www.wikidata.org/wiki/Q87748614)
  - [WikiProject Medicine](https://www.wikidata.org/wiki/Q4099686)

**Which of the existing aspects could or should link to or from this new aspect?**

- Essentially any of the other aspects could have a panel "relevant WikiProjects"

**Anything else?**

While I was using the P5008/QID combination as an example, I think this could also be implemented via a dedicated ```/wikiproject``` aspect. However, I think a generic PID/QID aspect would be useful more generally.

Daniel-Mietchen:
While working on this, we should keep in mind that WikiProjects are drifting towards using [P6104](https://scholia.toolforge.org/P6104) (maintained by WikiProject) versus [P5008](https://scholia.toolforge.org/P5008), but this transition is by no means finished, formalized or mandatory. So we should handle both flexibly.

PR #2289, opened 2023-06-15
Add WikiProject aspect

carlinmack:
Created a new aspect for WikiProjects. Has panels for items maintained by the WikiProject and items on the focus list of the WikiProject

image

I have also added a new function `get_prop_value` to `query.py`, which for a given QID and property, returns the value of the (Q, property, value) (where the value is also a QID. This is used to return the main subject.

### Caveats

* [x]  My code passes the [tox](https://tox.readthedocs.io/en/latest/) check, I can receive warnings about tests, documentation or both

### Testing

* If there are no items maintained by the [wikiproject](https://www.wikidata.org/wiki/Q18437553):

image

### Checklist

CC @Daniel-Mietchen


carlinmack:
Will consider adding panels suggested here https://github.com/WDscholia/scholia/issues/1853

Daniel-Mietchen:
> Will consider adding panels suggested here #1853

Yes, please do. Note that the second panel would give better results for P6104 instead of P5008, as per [this comment](https://github.com/WDscholia/scholia/issues/1853#issuecomment-1593184563).

carlinmack:
Daniel, if you look at the image you will see that I already use both P6104 and P5008 :)

carlinmack:
Once Shex schemas are queryable via SPARQL ([coming soon](https://www.wikidata.org/wiki/Wikidata_talk:Schemas#New_EntitySchema_data_type_is_ready_for_testing_on_Test_Wikidata)) I'll add a panel which shows the coverage of the schema properties for the relevant items.

carlinmack:
Added panels from a selection of the panels of the topic aspect:

image
image

scholia/query.py
```diff
@@ -295,6 +295,54 @@ def identifier_to_qs(property, identifier):
...
+        Property
+
+    Returns
+    -------
+    v : str
+        qid of value
+
+    Notes
+    -----
+    The Wikidata Query Service is queries to resolve the given identifier. If
```
fnielsen, line 315: queries -> queried

scholia/app/templates/wikiproject.html
```diff
@@ -0,0 +1,109 @@
...
+
+
+var url = 'https://www.wikidata.org/w/api.php?action=wbgetentities&ids=' +
+    '{{ main_subject }}' +
+    '&format=json&callback=?';
+  
+$.getJSON(url, function (data) {
+  var item = data.entities["{{ main_subject }}"];
+  if ('en' in item.labels) {
+    $("#main-subject").append(': <a href="/topic/{{ main_subject }}">' + item.labels.en.value + '</a>');
```
fnielsen, line 30: Do we need an escape here for item.labels.en.value?

carlinmack:
Had a call with Daniel where we worked on this aspect and he suggested using P6104/P5008 maintained by/focus list rather than getting the P921 main subject for the item. I think I'll remove get_prop_val in this case

fnielsen:
Let me know when you are finished.

carlinmack:
Removed the unused function, now ready to merge :) @fnielsen

fnielsen:
https://scholia.toolforge.org/wikiproject/Q56241615

PR #2292, opened 2023-07-01
Add co-occurring WikiProject panel

carlinmack:
Daniel mentioned that it would be useful to see co-occurring WikiProjects like we do with co-occurring topics on the topics aspect

#### WikiProject Invasion Biology    
image

#### WikiProject Covid-19
image

#### WikiProject Film
image

### Caveats

### Testing

* Test A
* Test B

### Checklist

### DIFF
--- a/scholia/app/templates/wikiproject.html
+++ b/scholia/app/templates/wikiproject.html
@@ -6,7 +6,7 @@
 
 {{ sparql_to_table("maintained") }}
 {{ sparql_to_table("focus") }}
-{{ sparql_to_table('co-occurring-wikiprojects') }}
+{{ sparql_to_table('co-occurring') }}
 
 {{ sparql_to_iframe('context') }}
 {{ sparql_to_table('recently-published-works') }}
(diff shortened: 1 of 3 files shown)


scholia/app/templates/wikiproject.html
```diff
@@ -6,6 +6,7 @@
 
 {{ sparql_to_table("maintained") }}
 {{ sparql_to_table("focus") }}
+{{ sparql_to_table('co-occurring') }}
```
fnielsen, line 9: This id does not correspond to the id in the HTML tag.

fnielsen:
https://scholia.toolforge.org/wikiproject/Q115206846#co-occurring-wikiprojects

fnielsen:
I changed the filename back to what you originally had. Would this not be the most "natural"?

carlinmack:
I was matching https://github.com/WDscholia/scholia/blob/master/scholia/app/templates/topic_co-occurring.sparql

fnielsen:
Ok. I see. I do not know whether we should change it then. Currently it fits with the headline, but there is inconsistency.

PR #2307, opened 2023-07-22
WDQS-compatible titles

### LINKED ISSUE #2304: is the `# title` in the SPARQL queries meant to be used by the WDQS?
## Context
Multiple SPARQL queries have a title in a comment, e.g. https://github.com/WDscholia/scholia/blob/6d14811f1a52ec68c2b359f64807b306fd82d257/scholia/app/templates/venue_topics.sparql#L1

But many of `# title:` which is not recognized by the WDQS, while `#title:` is (see below screenshots).

The space is [quite prevalent](https://github.com/search?q=repo%3AWDscholia%2Fscholia+%22%23+title%22+language%3ASPARQL&type=code&l=SPARQL).

## Question

Is the use of `#title:` in the queries meant to be used by the WDQS?

## Screenshots

With space:

image

Without space:

image

egonw:
Fixes #2304

### Description
Removes the space so that the `#title:` information can be used by the WDQS.
    
### Caveats
While not aware of it, I am not 100% sure there is not a downstream use of the space.

* [x]  Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Testing
* See the example in https://github.com/WDscholia/scholia/issues/2304

### Checklist

### DIFF
--- a/scholia/app/templates/chemical_identifiers.sparql
+++ b/scholia/app/templates/chemical_identifiers.sparql
@@ -1,6 +1,6 @@
 PREFIX target: <http://www.wikidata.org/entity/{{ q }}>
 
-# title: identifiers for this chemical
+#title: identifiers for this chemical
 SELECT
   ?Identifier ?IdentifierLabel
   ?Value (SAMPLE(?idUrls) as ?ValueUrl)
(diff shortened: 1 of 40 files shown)


fnielsen:
The "title" keyword was appropriated by the Wikidata developers it seems. If we start using it, then we end up with an extra header. Perhaps we could use the word "#description" instead ...

PR #2329, opened 2023-08-16
Implement the new chemical entity structure

egonw:
See https://www.wikidata.org/wiki/Wikidata:WikiProject_Chemistry/Guidelines/Basic_metaclasses_and_relations#Basic_metaclasses_for_chemical_entities

This is urgent because the use of 'chemical compound' is being phased out, and most chemical compounds would map to the `/chemical-class/` aspect instead of the correct `/chemical/` aspect.

### Description
This patch updates the matching by Scholia to the correct chemical aspects. It also updates the label of two classes.
    
### Caveats

* [x]  This change requires a documentation update
    * [x]  I have made corresponding changes to the documentation

* [x]  My code passes the [tox](https://tox.readthedocs.io/en/latest/) check, I can receive warnings about tests, documentation or both

### Testing
The following CLI call should give the following replies:

```shell
(scholia) ~/var/Projects/hub/scholia$ python -m scholia.query q-to-class Q899008
chemical
(scholia) ~/var/Projects/hub/scholia$ python -m scholia.query q-to-class Q425061
chemical_class
(scholia) ~/var/Projects/hub/scholia$ python -m scholia.query q-to-class Q411618
chemical_class
```

### Checklist

### DIFF
--- a/scholia/query.py
+++ b/scholia/query.py
@@ -1211,6 +1211,7 @@ def q_to_class(q):
             'Q79529',  # chemical substance
             'Q407595',  # metabolite
             'Q2393187',  # molecular entity
+            'Q113145171',  # type of a chemical entity
             ]):
         class_ = 'chemical'
     elif set(classes).intersection([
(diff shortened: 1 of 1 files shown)


fnielsen:
I saw this as urgent and have push it to the web servic: https://scholia.toolforge.org/Q46995757

fnielsen:
I should be getting more time in the end of August to do the PR in line.

egonw:
I still have some stuff to put in the paper, so that should work out. Each time I start writing there is more I "need" to do in Wikidata or on the Scholia code :/

PR #2342, opened 2023-09-24
Add ability to find new papers to add to Wikidata via topic curation pages

carlinmack:
Related to #1855

**Merge #2337 and #2340 first**

Adds support for searching Crossref for the 20 most recent papers on a disease, taxon, gene or protein and sends these results to the tool I've developed to find their presence in Wikidata and creating Quickstatements if they are found

image

image

#### Roadmap:

#2337 Adding support for multiple IDs
#2340 Adding support for DOIs to Quickstatements
This PR: Adding a button to curation pages to find if there are recent papers which can be created on Wikidata

Next PRs:

* Upgrading citations.js (haven't been able to figure out) and potentially contributing some stuff upstream
* Ability to select which entries to copy/send to quickstatements
* Add checking whether Wikidata entries can be enriched
    
### Caveats

> Please list anything which has been left out of this PR or which should be considered before this PR is accepted
Check any of the following which apply:

* [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ]  This change requires a documentation update
    * [ ]  I have made corresponding changes to the documentation
* [ ]  This change requires new dependencies (please list)

*If you make changes to the Python code*
  
* [ ]  My code passes the [tox](https://tox.readthedocs.io/en/latest/) check, I can receive warnings about tests, documentation or both

### Testing
Please describe the [tests](https://numpy.org/doc/stable/reference/testing.html) that you ran to verify your changes. Provide instructions, so we can reproduce. Please also list any relevant details for your test configuration.

* Test A
* Test B

### Checklist
* [ ] I have [commented](https://peps.python.org/pep-0008/#comments) my code, particularly in hard-to-understand areas
* [ ] My changes generate no new warnings
* [ ] I have not used code from external sources without attribution
* [ ] I have considered [accessibility](https://www.w3.org/standards/webdesign/accessibility) in my implementation 
* [ ] There are no remaining debug statements (print, console.log, ...)

PR #2340, opened 2023-09-18
Add support for DOI to QuickStatements

carlinmack:
Builds on top of #2337. **Merge #2337 before this PR**

#### Roadmap

#2337: Add support for multiple ArXiv IDs

This PR: Add support for other identifiers (DOI for now) and change name

#2342: Add new tool which allows users to query for most recent papers which will be sent to this tool for finding presence in Wikidata and creating Quickstatements (CC @Daniel-Mietchen)

PR I'd like to do if I can: 
* Upgrading citations.js (haven't been able to figure out) and potentially contributing some stuff upstream
* Ability to select which entries to copy/send to quickstatements
* Add checking whether Wikidata entries can be enriched

Introduces DOI to QuickStatements (using the Crossref API) resulting in renaming the tool to Identifier to QuickStatements. 
    
image

Note that this does not affect the current `/doi/[doi]` tool which creates QuickStatements using Citations.js and has different behaviour.

### Caveats

> Please list anything which has been left out of this PR or which should be considered before this PR is accepted
Check any of the following which apply:

* [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ]  This change requires a documentation update
    * [ ]  I have made corresponding changes to the documentation
* [ ]  This change requires new dependencies (please list)

*If you make changes to the Python code*
  
* [ ]  My code passes the [tox](https://tox.readthedocs.io/en/latest/) check, I can receive warnings about tests, documentation or both

### Testing
Please describe the [tests](https://numpy.org/doc/stable/reference/testing.html) that you ran to verify your changes. Provide instructions, so we can reproduce. Please also list any relevant details for your test configuration.

* I tested using identifiers found in Wikidata and not
* I created a test in test_utils.py to test the changes I made to `string_to_type`

### Checklist
* [ ] I have [commented](https://peps.python.org/pep-0008/#comments) my code, particularly in hard-to-understand areas
* [ ] My changes generate no new warnings
* [x] I have not used code from external sources without attribution
* [x] I have considered [accessibility](https://www.w3.org/standards/webdesign/accessibility) in my implementation 
* [x] There are no remaining debug statements (print, console.log, ...)

PR #2337, opened 2023-09-16
Add support for multiple ArXiv IDs

carlinmack:
#### Roadmap

This PR: Add support for multiple ArXiv IDs (in progress)

#2340 : Add support for other identifiers (doi for now) and change name

#2342: Add new tool which allows users to query for most recent papers which will be sent to this tool for finding presence in Wikidata and creating Quickstatements (CC @Daniel-Mietchen)

PR I'd like to do if I can: Upgrading citations.js (haven't been able to figure out) and potentially contributing some stuff upstream 

-----------------------------------------

Adds ability to submit multiple ArXiv IDs to the tool

image
    
### Caveats

> Please list anything which has been left out of this PR or which should be considered before this PR is accepted
Check any of the following which apply:

* [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ]  This change requires a documentation update
    * [ ]  I have made corresponding changes to the documentation
* [ ]  This change requires new dependencies (please list)

*If you make changes to the Python code*
  
* [ ]  My code passes the [tox](https://tox.readthedocs.io/en/latest/) check, I can receive warnings about tests, documentation or both

### Testing
Please describe the [tests](https://numpy.org/doc/stable/reference/testing.html) that you ran to verify your changes. Provide instructions, so we can reproduce. Please also list any relevant details for your test configuration.

* Tested with IDs which and found and not found (none, single and multiple)
* Added a warning if there is an exception/no quickstatements are generated
image

### Checklist
* [ ] I have [commented](https://peps.python.org/pep-0008/#comments) my code, particularly in hard-to-understand areas
* [x] My changes generate no new warnings
* [x] I have not used code from external sources without attribution
* [x] I have considered [accessibility](https://www.w3.org/standards/webdesign/accessibility) in my implementation 
* [x] There are no remaining debug statements (print, console.log, ...)

PR #2346, opened 2023-10-08
Create Check Crossref tool

carlinmack:
Builds on top of #2337 and #2340 Merge these first

This is the same functionality provided by the curation pages in #2342 but allows the user to edit the query directly rather than using the defaults set by the curation page tool

Users can either use the simple form inputs or edit the query string directly in the Advanced section. The "Submit" button sends the DOIs to the ID to QS tool, "Test query" button allows users to see the API results directly
    
image

Note: the date formatting in the date picker is set with the user's locale setting. For example in another of my browsers, the date shows as YYYY-MM-DD

### Caveats

### Testing

* Checked that everything works as expected

### Checklist


fnielsen:
https://scholia.toolforge.org/check-crossref

PR #2358, opened 2023-10-20
more chemical-class updates for the updated Wikidata chem hierarchy model

egonw:
Fixes various issues, including the visual hierarchy plot. Continues the patch in another panel: https://github.com/WDscholia/scholia/pull/2344

### Description
Previously, specific chemicals where `instance of` a chemical class, but they are now always `subclass of`. To continue the differences in the `chemical-class` aspect, we need to do this check differently. Some updates do this by checking if the subclass itself has subclasses, mimicking the `P31` behavior before. In other places, the `P31` can just be removed because it used "one or more". For those panels, we do not see differences.
    
### Caveats

    * [x]  I have made corresponding changes to the documentation

### Testing
After the patch, some of the nodes should be orange again:

* https://scholia.toolforge.org/chemical-class/Q41581

### Checklist

### DIFF
--- a/scholia/app/templates/chemical-class_publications-per-year.sparql
+++ b/scholia/app/templates/chemical-class_publications-per-year.sparql
@@ -12,7 +12,7 @@ select ?year (count(?work) as ?number_of_publications) where {
       bind(year(?date) as ?year_)
       {
         select (min(?year_) as ?earliest_year) where {
-          { ?work wdt:P921/wdt:P31*/wdt:P279* target: . }
+          { ?work wdt:P921/wdt:P279* target: . }
           union { ?work wdt:P921/wdt:P361+ target: . }
           union { ?work wdt:P921/wdt:P1269+ target: . }
           union { target: ?propp ?statement .
@@ -28,7 +28,7 @@ select ?year (count(?work) as ?number_of_publications) where {
   }
   union {
     select ?work (min(?years) as ?year) where {
-      { ?work wdt:P921/wdt:P31*/wdt:P279* target: . }
+      { ?work wdt:P921/wdt:P279* target: . }
       union { ?work wdt:P921/wdt:P361+ target: . }
       union { ?work wdt:P921/wdt:P1269+ target: . }
       union { target: ?propp ?statement .
--- a/scholia/app/templates/chemical-class_related-chemicals.sparql
+++ b/scholia/app/templates/chemical-class_related-chemicals.sparql
@@ -2,7 +2,8 @@ PREFIX target: <http://www.wikidata.org/entity/{{ q }}>
 
 SELECT ?mol ?molLabel ?InChIKey ?CAS ?CASUrl ?ChemSpider ?ChemSpiderUrl ?PubChem_CID ?PubChem_CIDUrl WITH {
 SELECT DISTINCT ?mol WHERE {
-  ?mol wdt:P31/wdt:P279* target: .
+  ?mol wdt:P279* target: .
+  MINUS { [] wdt:P279 ?mol }
 } LIMIT 500
 } AS %result
 WHERE {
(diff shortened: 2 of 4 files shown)


scholia/app/templates/chemical-class_related-chemicals.sparql
```diff
@@ -2,7 +2,8 @@ PREFIX target: <http://www.wikidata.org/entity/{{ q }}>
 
 SELECT ?mol ?molLabel ?InChIKey ?CAS ?CASUrl ?ChemSpider ?ChemSpiderUrl ?PubChem_CID ?PubChem_CIDUrl WITH {
 SELECT DISTINCT ?mol WHERE {
-  ?mol wdt:P31/wdt:P279* target: .
+  ?mol wdt:P279* target: .
+  MINUS { [] wdt:P279 ?mol }
```
fnielsen, line 6: Could these two lines not be changed to `?mol wdt:P279+ target:`?

scholia/app/templates/chemical-class_publications-per-year.sparql
```diff
@@ -12,7 +12,7 @@ select ?year (count(?work) as ?number_of_publications) where {
       bind(year(?date) as ?year_)
       {
         select (min(?year_) as ?earliest_year) where {
-          { ?work wdt:P921/wdt:P31*/wdt:P279* target: . }
+          { ?work wdt:P921/wdt:P279* target: . }
           union { ?work wdt:P921/wdt:P361+ target: . }
           union { ?work wdt:P921/wdt:P1269+ target: . }
```
fnielsen, line 17: Should we change the 3 lines to `wdt:P921 / (wdt:P31 | wdt:P279 | wdt:P361 | wdt:P1269)* target: ;

scholia/app/templates/chemical-class_related-chemicals.sparql
egonw: The `MINUS` is in the other direction.

The `?mol wdt:P279* target: .` could be `?mol wdt:P279+ target: .`, but do you think that makes a difference? What would it match for no `wdt:P279` ? (I do not actually know...)

scholia/app/templates/chemical-class_publications-per-year.sparql
egonw: I am not sure... how does ``wdt:P921 / (wdt:P31 | wdt:P279 | wdt:P361 | wdt:P1269)* target:` expand? Does that include mixing?

scholia/app/templates/chemical-class_related-chemicals.sparql
fnielsen: Will `?mol wdt:P279* target:` not match target=?mol ?

scholia/app/templates/chemical-class_publications-per-year.sparql
fnielsen: It might be too expensive. In Synia I started to use ` ?work wdt:P921 / (wdt:P31 | wdt:P279 | wdt:P361 | wdt:P1269)? # Expensive: *
            target: ;`  here https://synia.toolforge.org/#topic/Q515701

scholia/app/templates/chemical-class_related-chemicals.sparql
egonw: I'll check.

scholia/app/templates/chemical-class_publications-per-year.sparql
fnielsen: I know we have the UNION-like way all over, but to me "wdt:P921 / (wdt:P31 | wdt:P279 | wdt:P361 | wdt:P1269)*" looks prettier, e.g., why is there a P279 for P31 but not the others?

scholia/app/templates/chemical-class_related-chemicals.sparql
fnielsen: Isn't MINUS there to avoid that `?mol` becomes `target:`?

egonw: No. The MINUS is to ensure it only colors things orange that are not a superclass themselves.

PR #2363, opened 2023-10-20
Fix finding taxons in which this class is found

egonw:
Fixes listing the taxons in which chemicals in this chemical class are found.

Also removes an unused OPTIONAL part.

### Description
The list was empty because it still used the P31 structure. Similar to other recent fixes.
    
### Caveats

### Testing
* Got to https://scholia.toolforge.org/chemical-class/Q2832210
* Before the patch the list is empty
* After the patch it should look like this:

image

### Checklist


fnielsen:
The is commited by https://github.com/WDscholia/scholia/commit/8dce24f64e22a7784b0dbb2c93e11323ab185215

PR #2382, opened 2023-11-14
Count all unique items with a P31

egonw:
Hi Finn, this is a question if an enhancement list this makes sense. It's doesn't seem to slow down the query and answers the worry that articles are the major part of the Wikidata items.


fnielsen:
Is the description correct? What is a Thing? Couldn't a Thing be a class without P31? Why not "instances"?

egonw:
Yeah, I like the idea of "Number of instances"!

fnielsen:
I merge it anyway :)

egonw:
> I merge it anyway :)

As gatekeeper, that's totally up to you :)

fnielsen:
What is owl:Thing anyway? Would that be every item in Wikidata?

egonw:
"thing" even goed deeper, all the way to RDF, see https://www.w3.org/TR/rdf11-concepts/#resources-and-statements

PR #2393, opened 2023-12-11
Update award-index_statistics.sparql to make the graphic more straightforward

Adafede:
### Description

When looking at the current https://scholia.toolforge.org/award/#statistics, the first impression is that gender balance is "not so bad" as it seems there are quite some points above a "1/1 line", but this is heavily biased because the x and y axis do not share the same maximum.
This PR aims at fixing this.
    
### Caveats

Currently, the manual max value has been set to 2,000.

### Testing

* Test A
* Test B

### Checklist

### DIFF
--- a/scholia/app/templates/award-index_statistics.sparql
+++ b/scholia/app/templates/award-index_statistics.sparql
@@ -1,3 +1,4 @@
+# tool: scholia
 #defaultView:ScatterChart
 # Statistics on number of recipients for award wrt. gender
 SELECT ?males ?females ?award ?awardLabel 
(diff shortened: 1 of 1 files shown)


Adafede:
See https://w.wiki/8TqV (proposal) for comparison with https://w.wiki/8TqY (current)

Adafede:
Same strategy could be applied to some other graphics (e.g. https://scholia.toolforge.org/publisher/Q233358#citations-works, just landed on it randomly)

fnielsen:
Cherry-picked https://github.com/WDscholia/scholia/commit/974b55aa24dcee44c152d4302995bdace066ecee

PR #2399, opened 2024-01-17
Update Citation.js to 0.7.6

larsgw:
### Description
Update the Citation.js bundle to citation-js 0.7.6, as well as updates to various plugins. Bundle content as specified here: https://github.com/WDscholia/scholia/pull/1941. This is mostly to fix an issue with importing dissertations from Crossref (https://github.com/citation-js/citation-js/issues/220) but also includes other patches, including better BibTeX output.

### Checklist


larsgw:
In the future, I'd like to replace this with the [Toolforge API](https://citation-js.toolforge.org/) but it still gives some incomplete errors.

PR #2426, opened 2024-02-13
Ensure Article Titles with Dashes are Handled Appropriately for Identified Papers

### LINKED ISSUE #2420: OJS title extraction may differ between ":" and " - "
**Describe the bug**
OJS title extraction may differ between ":" and " - ". Scraping https://tidsskrift.dk/samfundsokonomen/article/view/143319 should give a title with ":" according to metadata on the page, but the quickstatements generated contain a form of dash: "Civilsamfund og velfærdsstat – konflikt, samarbejde eller begge dele?"

This is also the case with https://tidsskrift.dk/samfundsokonomen/article/view/143319

It means that the article is not match, - it is already in Wikidata, e.g., https://scholia.toolforge.org/work/Q123561283

**To Reproduce**
1. `python -m scholia.scrape.ojs issue-url-to-quickstatements https://tidsskrift.dk/samfundsokonomen/issue/view/10780`
2. `LAST	Len	"Civilsamfund og velfærdsstat – konflikt, samarbejde eller begge dele?"`

**Expected behavior**
Should identify https://scholia.toolforge.org/work/Q123561283

faresh9:
Fixes #2420

### Description
This PR modifies the code to ensure that article titles containing a dash are only replaced with a colon if the article is not identified. If the article is identified after the modification, the modified title is used; otherwise, the original title with the dash is retained.

### Caveats
No caveats.

### Testing
1. Test with an identified article containing a dash in the title.
2. Test with an unidentified article containing a dash in the title.
3. Verify that the title modification behaves as expected in both cases.

### Checklist
* [x] My code passes the tox check, I can receive warnings about tests, documentation, or both

### DIFF
--- a/scholia/scrape/ojs.py
+++ b/scholia/scrape/ojs.py
@@ -319,7 +319,6 @@ def scrape_paper_from_url(url):
     title = _fields_to_content(['citation_title', 'DC.Title',
                                 'DC.Title.Alternative'])
     if title is not None:
-        #original_title = title  # Store the original title
         q = paper_to_q(entry)
         if not q:  # If not identified before modification
             # Replace dash or colon without altering surrounding spaces
--- a/scholia/scrape/ojs.py
+++ b/scholia/scrape/ojs.py
@@ -319,7 +319,20 @@ def scrape_paper_from_url(url):
     title = _fields_to_content(['citation_title', 'DC.Title',
                                 'DC.Title.Alternative'])
     if title is not None:
-        entry['title'] = title
+        original_title = title  # Store the original title
+        q = paper_to_q(entry)
+        if not q:  # If not identified before modification
+            # Replace dash or colon without altering surrounding spaces
+            title = title.replace(" –", ":").replace("– ", ":")
+            entry['title'] = title
+            q = paper_to_q(entry)
+            if q:  # If identified after modification
+                entry['title'] = title  # Plug in the modified title
+            else:
+                entry['title'] = original_title  # Plug in the original title
+        else:
+            entry['title'] = original_title
+
 
     citation_date = _fields_to_content(['citation_date', 'DC.Date.issued'])
     if citation_date is not None:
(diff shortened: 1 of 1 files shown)


scholia/scrape/ojs.py
```diff
@@ -319,7 +319,20 @@ def _fields_to_content(fields):
     title = _fields_to_content(['citation_title', 'DC.Title',
                                 'DC.Title.Alternative'])
     if title is not None:
-        entry['title'] = title
+        original_title = title  # Store the original title
+        q = paper_to_q(entry)
+        if not q:  # If not identified before modification
+            # Replace dash or colon without altering surrounding spaces
+            title = title.replace(" –", ":").replace("– ", ":")
+            entry['title'] = title
```
fnielsen, line 327: If there is no replacement, then it is not necessary to look up the second time.
```
if orginal_title != title:
```
could be included to avoid bothering the API a second time, - in most cases.

faresh9: So i should check if the paper is identified or not just one time?

fnielsen: That depends, e.g., if the title is "An introduction to Scholia" and it does not exist in Wikidata, then doing a replacement would replace nothing (because there is no ":" or "-" in the title). So the second API lookup would return the same.

faresh9: Okay so if the title contains dash or colon the second API pass is okay ?

faresh9:
Everything should be good now

scholia/scrape/ojs.py
```diff
@@ -319,7 +319,23 @@ def _fields_to_content(fields):
     title = _fields_to_content(['citation_title', 'DC.Title',
                                 'DC.Title.Alternative'])
     if title is not None:
-        entry['title'] = title
+        original_title = title  # Store the original title
```
fnielsen, line 322: You have two variables `original_title` and `title` that contains the same. I do not think the `original_title` variable is necessary.

faresh9: i did it for extra clarification, but i removed it anyways

fnielsen:
There were styling errors that flake8 reported, therefore https://github.com/WDscholia/scholia/pull/2440

PR #2432, opened 2024-03-02
Added support for nanopublications as source

egonw:
### Description
This patch implements support for CiTO intention annotations from nanopublications. After a [discussion on Mastodon](https://social.edu.nl/@egonw/112011219884262416), there now is a [nanopublication template](https://nanodash.knowledgepixels.com/publish?2&template=https://w3id.org/np/RAX_4tWTyjFpO6nz63s14ucuejd64t2mK3IBlkwZ7jjLo&template-version=latest) for such annotation and a [SPARQL query](https://tinyurl.com/get-cito-nanopubs) can extract this and add this in an automated fashion to Wikidata.

I have [proposed to write a FAIR CookBook recipe](https://github.com/FAIRplus/the-fair-cookbook/issues/628).
    
### Caveats

None that I can think of. The runtime is still quite fast.

### Testing

Patch one: The query now adds a "Nanopublication" source (venue), as shown below after "unselecting" the other sources:

image

Patch two: this adds a `Number of nanopublications with explicit CiTO annotation` line to the statistics panel, like:

image

### Checklist

### DIFF
--- a/scholia/app/templates/cito-index_articles-by-year.sparql
+++ b/scholia/app/templates/cito-index_articles-by-year.sparql
@@ -3,12 +3,22 @@ select ?year (count(?work) as ?number_of_publications) ?role where {
   {
     select ?work (min(?years) as ?year) ?type_ ?venue where {
       ?work wdt:P577 ?dates ;
-            p:P2860 / pq:P3712 / wdt:P31 wd:Q96471816 .
+            p:P2860 ?citesStatement .
+      ?citesStatement pq:P3712 / wdt:P31 wd:Q96471816 .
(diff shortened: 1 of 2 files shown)


egonw:
BTW, the `tox` is because of an unrelated test fails, because the assumption was broken by an edits on the 29th, with a long batch of `DOI prefix` removal quickstatements: https://www.wikidata.org/w/index.php?title=Q463494&diff=2089683809&oldid=2077564810&diffmode=source

fnielsen:
Interestingly I get 0 for Number of nanopublications with explicit CiTO annotation when looking at http://127.0.0.1:8100/cito/

PR #2526, opened 2024-07-24
added birth and death dates to author timeline

stuartyeates:
Fixes # (issue number, if applicable) 

    
### Caveats

* I've only tested this on https://query.wikidata.org/ , I don't have a full local stack to test on. 
* This is my first contribution to this project and my first SPARQL contribution to any project.
* No changes to python code. Two new stanzas in SPQARL query.

### Testing

I ran the SPQARL query on https://query.wikidata.org/. the query is too long to use the URL shortener.

### Checklist
* [ X  ] My changes generate no new warnings
* [ X ] I have not used code from external sources without attribution
* [ X ] I have considered [accessibility](https://www.w3.org/standards/webdesign/accessibility) in my implementation 
* [ X ] There are no remaining debug statements (print, console.log, ...)

### DIFF
--- a/scholia/app/templates/author_timeline.sparql
+++ b/scholia/app/templates/author_timeline.sparql
@@ -94,5 +94,15 @@ WHERE {
     BIND(CONCAT("🏆 ",?award_label) AS ?label)
     ?award_statement pq:P585 ?beginning .
   }
-
+  UNION
+  {
+    target: wdt:P569 ?beginning .
+    BIND(CONCAT("👶 born ") AS ?label) .
+  }
+  UNION
+  {
+    target: wdt:P570 ?beginning .
+    BIND(CONCAT("💀 died") AS ?label) .
+  }
+			    
 } 


egonw (commented):
Great idea! I think it's the simplest approach, but for old authors very welcome.

Because it also shows up for people who died recently, sometimes young authors, I would suggest to not use emoiji's here. What do you think?

stuartyeates:
I'm not bothered with or without the skull. I tried the tombstone emoji, but it didn't work on any of the platforms I tested on so I figured it was too new to be useful.

fnielsen:
I almost recall some concert about showing birthdate and date of death. I suppose we can keep it for now and remove it if users find it inappropriate.

egonw:
> remove it if users find it inappropriate.

With the risk that those who find it inappropriate will just stay away from Scholia.

But I don't have strong arguments to remove it at this moment.

fnielsen:
ChatGPT suggests 🏺, ✝️, ⚰️, 🥀. The cross seems tied to a Christian culture, when reminded it suggest ⏳

fnielsen:
https://github.com/WDscholia/scholia/issues/2532

PR #2547, opened 2024-10-18
Demonstrate rendering author's publication list via SnapQuery

cthoyt:
Note, before you run this, you have to run `snapquery --import /Users/cthoyt/dev/snapquery/snapquery/samples/scholia.json` to import the scholia query definitions.

You can also add endpoint_name with one of `'wikidata', 'qlever-wikidata', 'qlever-dblp', 'qlever-dblp-plus', 'wikidata-openlinksw'` - http://127.0.0.1:8100/author/Q47475003?endpoint_name=wikidata-qlever. By default, if none is given, it uses `wikidata`

### DIFF
--- a/scholia/app/templates/author.html
+++ b/scholia/app/templates/author.html
@@ -59,7 +59,7 @@
     <td>{{ publication.date }}</td>
     <td><a href="{{ publication.work }}">{{ publication.workLabel }}</a></td>
     <td>{{ publication.type }}</td>
-    <td>{% if publication.pages %}{{ publication.pages|int }}{% endif %}</td>
+    <td align="right">{% if publication.pages %}{{ publication.pages|int }}{% endif %}</td>
     <td><a href="{{ publication.venue }}">{{ publication.venueLabel }}</a></td>
     <td><a href="{{ publication.authorsUrl }}">{{ publication.authors }}</a></td>
 </tr>
--- a/scholia/app/templates/author.html
+++ b/scholia/app/templates/author.html
@@ -4,7 +4,6 @@
 
 {% block in_ready %}
 
-{{ sparql_to_table('list-of-publications') }}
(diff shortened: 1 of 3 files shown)


egonw:
It's a WIP, so you're probably already thinking about it, but we need to rework this line:

https://github.com/WDscholia/scholia/pull/2547/files#diff-1332f7fcd25ac819245542532ec4dc747b61257c03c93541246a034918f07616L7

And have the table dynamically created here too.

Thinking out loud. What exactly do we win with this? Because snapquery still needs separate SPARQL queries for different backends. Also, we likely have to setup snapquery on toolforge, right?

fnielsen:
* Does this relates to an issue?
* There is server-side render (author.html). I would be afraid of this. 
* Is it not not asynchronous (synchronous) (views.py, line 406)?
* The setup of a panel requires coordination of the HTML table setup and the "remote" named SPARQL query.

egonw:
@fnielsen, can you explain to @cthoyt what kind of change you expect to be able to consider this patch? This is not really clear to me.

fnielsen:
The problem is that there server is locked during the call to the the bitplan SPARQL endpoing as far as I can see. The formating of the table should happen on the client-side so we are not overloading Toolforge. We have on average around 3-4 requests per second and four servers running Scholia, so there around 1 request per second on average. We should get as quickly as possible out of the server, i.e., return a response as fast as we can. A network request to, e.g., bitplan would probably take several 100 ms, so we might quickly congest the server.

Furthermore, it is not clear to me that we should use an Wikimedia-external server - if this is implemented client-side.

So far we have formatted the table based on the SPARQL only. And there is only one file we need to edit if the table is changed. This PR diverge from this approach.

It is difficult to me to see how we can change this patch to so it fits into our common Scholia approach.

PR #2589, opened 2025-04-02
update to have Scholia use the legacy SPARQL endpoint

egonw:
"Fixes" not using the full data. 

This is not a really a fix, but a workaround offered to us by Wikimedia. cc @bluerasberry


fnielsen:
It now uses query-legacy-full. https://scholia.toolforge.org/work/Q131192529

egonw:
Thanks!

PR #2597, opened 2025-04-06
Added titles for the chemistry aspect panels

egonw:
Adds missing query titles, needed for downstream use.

### Description
Not all `chemical` aspect queries have titles:

image

This patch adds a good number of them.
    
### Caveats
The titles are listed as comments in the SPARQL. If not well-formed, it can break running the SPARQL query.

### Testing
Check if the following aspects still show content in all panels:

* `chemical`
* `chemical-class`
* `chemical-element`

### Checklist

### DIFF
--- a/scholia/app/templates/chemical-class_class-hierarchy.sparql
+++ b/scholia/app/templates/chemical-class_class-hierarchy.sparql
@@ -1,3 +1,4 @@
+# title: network depiction of the class hierarchy
 #defaultView:Graph
 PREFIX target: <http://www.wikidata.org/entity/{{ q }}>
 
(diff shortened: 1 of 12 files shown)


fnielsen:
Isn't is a problem that WDQS intepretes `#title` as something? When we rely on the whitespace could that not go run?

Daniel-Mietchen:
@egonw Can you adjust your workflows (and thus the PR) such that there is no space between the ```#``` and the ```title:```? 
Without this space, the titles would be compatible with the way (Wikimedia's version of) Blazegraph handles titles.

egonw:
@Daniel-Mietchen, @fnielsen, I left the space deliberately, per https://github.com/WDscholia/scholia/issues/2304#issuecomment-1653899623

PR #2611, opened 2025-04-12
Lazy loading of table and iframe panels

### LINKED ISSUE #2602: Avoid concurrent SPARQL queries while loading a Scholia page
**Is your feature request related to a problem? Please describe.**

For Scholia pages with more than one SPARQL query, we're currently (since the switch to the legacy endpoint) more frequently getting the error message 

> This query has timed out, we recommend that you follow the link to the Wikidata Query Service below to modify the query to be less intensive. 

as shown below for the Identifiers section of https://scholia.toolforge.org/taxon/Q202864 .

Image

This is likely due to concurrent querying of the same endpoint server, since the current setup only uses one server, whereas the previous one had about a dozen.

**Describe the solution you'd like**
We would add some lazy loading, promises or similar that reduce the number of concurrent queries and thus the frequency of these error messages.

**Describe alternatives you've considered**
It would be nice to have multiple machines supporting the legacy endpoint, but that is currently not an option.

**Additional context**
The graph split resulted in a switch of Scholia's SPARQL endpoint, as per https://www.wikidata.org/wiki/Wikidata:Scholia/Events/Hackathon_April_2025 .

egonw:
Fixes #2602

### Description
The patch uses the `Intersection Observer API`. For each panel, it creates a unique observer. When a panel becomes visible, the (unique) handler is called. This is followed by an unobserve to ensure that the SPARQL query is run only once.
    
### Caveats
This browser compability page shows that the `Intersection Observer API` is widely available, but not on old browsers: https://caniuse.com/intersectionobserver

* [x]  Breaking change (unsupported for old browsers)

### Testing
Open up as many pages as possible and see if you see irregularities.

### Checklist

### DIFF
--- a/scholia/app/templates/base.html
+++ b/scholia/app/templates/base.html
@@ -12,13 +12,13 @@ options={{ options | tojson }});
 
 {% macro sparql_to_table(panel, options={}) -%}
 // {{ panel }} table
-console.log('adding an observer to: #{{ panel }}-table');
+// console.log('adding an observer to: #{{ panel }}-table');
 const {{ panel | replace("-", "") }}Observer = new IntersectionObserver({{ panel | replace("-", "") }}HandleIntersection);
 {{ panel | replace("-", "") }}Observer.observe(document.querySelector('#{{ panel }}-table'));
 function {{ panel | replace("-", "") }}HandleIntersection(entries, observer){
   entries.forEach(entry => {
     if (entry.intersectionRatio > 0) {
-      console.log('Element is in the viewport! #{{ panel }}-table');
+      // console.log('Element is in the viewport! #{{ panel }}-table');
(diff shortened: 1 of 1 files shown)


Daniel-Mietchen:
This seems to work fine, and the console message comments can probably be removed.

However, I am still getting occasional cases where I am running into a rate limit (usually by opening multiple Scholia pages in short succession, e.g. by clicking through the example links at aspect landing pages like https://scholia.toolforge.org/organization/ )

grafik

taken from https://scholia.toolforge.org/organization/Q1269766#page-production .

In such cases, the Reload button does not appear. Shall I open a new ticket for that?

egonw:
@fnielsen, can you please let me know if this patch is good to go?

Issue #2668, opened 2025-06-09
WDQS server names displayed without spaces


egonw:
**Describe the bug**
The WDQS name is not shown correctly [when using the new config](https://github.com/WDscholia/scholia/pull/2640), and spaces seem to get removed: 

![Image](https://github.com/user-attachments/assets/572fb040-039f-4bea-ae62-d89982be5ec7)

**To Reproduce**
Steps to reproduce the behavior. Update any SPARQL query with the following SPARQL-level config:

```
#sparql_endpoint = https://query.wikidata.org/sparql
#sparql_endpoint_name= WDQS main
#sparql_editurl = https://query.wikidata.org/#
#sparql_embedurl = https://query.wikidata.org/embed.html#
```

Note the name in the second line, which has the space.

Look up the aspect with the SPARQL query you changes, run the local server, and see how the space is dropped. (Also with multiple spaces in the EP name)

**Expected behavior**
The UI should include the space, as given in the config.

Issue #2667, opened 2025-06-09
`chemical-class` SPARQL queries need checking to not run on legacy


egonw:
The follow panel/SPARQL queries need updating:

- [ ] https://github.com/WDscholia/scholia/blob/main/scholia/app/templates/chemical-class_class-hierarchy.sparql
- [ ] https://github.com/WDscholia/scholia/blob/main/scholia/app/templates/chemical-class_identifiers.sparql
- [ ] https://github.com/WDscholia/scholia/blob/main/scholia/app/templates/chemical-class_related-chemicals.sparql
- [ ] https://github.com/WDscholia/scholia/blob/main/scholia/app/templates/chemical-class_recent-literature.sparql
- [ ] https://github.com/WDscholia/scholia/blob/main/scholia/app/templates/chemical-class_publications-per-year.sparql
- [ ] https://github.com/WDscholia/scholia/blob/main/scholia/app/templates/chemical-class_found-in-taxon.sparql

PR #2669, opened 2025-06-09
Checked two `chemical-class` queries and they can be run on WDQS main

egonw:
Fixes part of https://github.com/WDscholia/scholia/issues/2667

The two queries do not use scholarly information.

### Description
Comparison of the Legacy (right) versus Main (left):

image

### Caveats
The pull request is affected by this bug: https://github.com/WDscholia/scholia/issues/2668

### Testing
Run the two queries on Legacy and Main. The test fail seems to be due to an unrelated tox test assertion.

For example, when locally testing `/chemical-class/Q41581`, check if you see the two panels use `query.wikidata.org` instead of `query-legacy-full.wikidata.org`:

image

### Checklist

### DIFF
--- a/scholia/app/templates/chemical-class_identifiers.sparql
+++ b/scholia/app/templates/chemical-class_identifiers.sparql
@@ -1,4 +1,8 @@
 # title: identifiers for this chemical class
+#sparql_endpoint = https://query.wikidata.org/sparql
+#sparql_endpoint_name= WDQS main
+#sparql_editurl = https://query.wikidata.org/#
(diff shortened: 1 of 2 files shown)


fnielsen:
> There are failing tests.

Now they are working.

egonw:
> Now they are working.

@fnielsen, what was the problem? as far as I can tell, the problems were not caused by this path, or were they?

fnielsen:
@egonw Sometimes the test top because of an API call does not work. It usually helps to redo the test by pressing the test button manually.

PR #2715, opened 2025-09-04
Update of the `main` branch to work with QLever

hannahbast:
@fnielsen, this is @egonw typing here, this PR started out as "Configure Flask app for proxy deployment" but now has many updates to make Scholia work with QLever. The live instance is shaping up really well. But there is so much here, plz ping me to talk you through the work.

@hannahbast, sorry for updating the title and this description. Since everything was merged into the `qlever` branch, it all shows up on this PR.


egonw:
I wonder if the proxy info can be integrated in the configuration system we have, instead of having it hardcoded.

fnielsen:
I am currently swarm in GitHub messages and have some other obligations.

